PR #1, opened 2018-10-12
Replace jquery global with promises

epeach:
Removes dependence on JQuery $

### DIFF
--- a/src/p5.dance.js
+++ b/src/p5.dance.js
@@ -631,7 +631,9 @@ export default class DanceParty {
 
   loadSongMetadata_(id){
     let songDataPath = '/api/v1/sound-library/hoc_song_meta';
-    return new Promise((resolve) => fetch(`${songDataPath}/${id}.json`).then((data) => {METADATA[id] = data; resolve()}));
+    return new Promise((resolve) => fetch(`${songDataPath}/${id}.json`)
+      .then((response) => {return response.json()})
+      .then((data) => {METADATA[id] = data; resolve()}));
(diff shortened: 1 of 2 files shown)


src/p5.dance.js
```diff
@@ -629,20 +629,16 @@ export default class DanceParty {
     return this.p5_.allSprites.indexOf(sprite) > -1;
   }
 
-  loadSongMetadata_(callback) {
+  loadSongMetadata_(id){
     let songDataPath = '/api/v1/sound-library/hoc_song_meta';
+    return new Promise((resolve) => fetch(`${songDataPath}/${id}.json`).then((data) => {METADATA[id] = data; resolve()}));
```
joshlory: I think `data` is actually a response object, and to get the actual text we need to do `data.text()` which returns a promise.

Small nit, we can use `await` here:

```js
  /** @return Promise */
  async loadSongMetadata_(id) {
    const response = await fetch(`${songDataPath}/${id}.json`);
    METADATA[id] = await response.text();
  };
```

joshlory: Or if you want to defend against a user spamming the "Run" button, and return the pending promise instead of creating a new one:

```js
  /** @return Promise */
  async loadSong_(path) {
    if (!this.loaders_[path]) {
      this.loaders_[path] = await fetch(path);
    }
    return await this.loaders_[path].text();
  };
```

epeach:
@joshlory PTAL, thanks!

PR #2, opened 2018-10-12
Add group creation, property setting, and additional layouts

mrjoshida:
Add commands for creating a group of anonymous sprites, setting group properties, and new layouts (circle and random).

### DIFF
--- a/src/p5.dance.js
+++ b/src/p5.dance.js
@@ -390,7 +390,7 @@ export default class DanceParty {
       var radius = 50 + (count * 5);
       var angle = -90 * (Math.PI / 180);
       var step = (2 * Math.PI) / count;
-      group.forEach(function(sprite) {
+      group.forEach(function (sprite) {
         sprite.x = 200 + (radius * Math.cos(angle));
         sprite.y = 200 + (radius * Math.sin(angle));
         angle += step;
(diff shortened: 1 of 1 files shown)


src/p5.dance.js
```diff
@@ -299,6 +299,14 @@ export default class DanceParty {
     return sprite;
   }
 
+  makeNewDanceSpriteGroup(n, costume, layout) {
```
joshlory: Should we put a cap on `n`?

src/p5.dance.js
```diff
@@ -403,12 +423,17 @@ export default class DanceParty {
...
         sprite.y = 200;
       }
-    } else {
+    } else if (format === "column") {
       for (i=0; i<count; i++) {
         sprite = group[i];
         sprite.x = 200;
         sprite.y = (i+1) * (400 / (count + 1));
       }
+    } else if (format == "random") {
```
joshlory: Nit: `===`

src/p5.dance.js
mrjoshida: I'm capping it in the block by providing a dropdown (currently set to 3-10)

joshlory: We probably want to also cap it here (or in Dance.js `nativeAPI. makeNewDanceSpriteGroup` call).  A user can hand-edit the XML to bypass any dropdown limits.

joshlory:
image

Rad! `nativeAPI.makeNewDanceSpriteGroup(10, 'MOOSE', 'circle');`

src/p5.dance.js
mrjoshida: Does it make more sense to cap the total number of sprites across the board instead. Even if we cap this block, you could just run it multiple times (which I would expect is a more common student workaround than editing XML)

mrjoshida: Or we let it be and see kids enjoy the Star Wars infinite droid effect :)

PR #3, opened 2018-10-12
Clean up metadata load

epeach:
Remove the 'songs' hard-coded meta-data and replaces with reference to metadata object.

### DIFF
--- a/src/p5.dance.js
+++ b/src/p5.dance.js
@@ -154,8 +154,7 @@ export default class DanceParty {
   }
 
   play() {
-    console.log("HERE: " + METADATA[this.getSelectedSong_()]);
-    this.playSound_({url: METADATA[this.getSelectedSong_()].url, callback: () => {this.songStartTime_ = new Date()}});
+    this.playSound_({url: METADATA[this.getSelectedSong_()].file, callback: () => {this.songStartTime_ = new Date()}});
   }
 
   setBackground(color) {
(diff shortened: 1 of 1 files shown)


src/p5.dance.js
```diff
@@ -20,37 +20,6 @@ const ANIMATIONS = {};
...
-    delay: 0.0, // Seconds to delay before calculating measures
-    verse: [0, 26.3], // Array of timestamps in seconds where verses occur
-    chorus: [65.75] // Array of timestamps in seconds where choruses occur
-  },
-  hammer: {
-    url: 'https://curriculum.code.org/media/uploads/touch.mp3',
-    bpm: 133,
-    delay: 2.32, // Seconds to delay before calculating measures
-    verse: [1.5, 15.2], // Array of timestamps in seconds where verses occur
-    chorus: [5.5, 22.1] // Array of timestamps in seconds where choruses occur
```
joshlory: Should the "verse" and "chorus" metadata be added to https://levelbuilder-studio.code.org/api/v1/sound-library/hoc_song_meta/hammer.json?

epeach: Good catch. I know JoshS is going to be working on finding these timestamps, so we'll probably have to find a better way to get it in metadata, but in the meantime, I'll just copy them over.

epeach: Updated the current S3 files with these values.

PR #4, opened 2018-10-13
Typo fix - Call GetCurrentFunction

epeach:
Adding dropped parentheses.


epeach:
Seems safe enough to merge without review. :)

PR #12, opened 2018-10-16
Add `tape` for basic unit tests

joshlory:
I'm following the pattern from the https://github.com/code-dot-org/craft repo, which seems to work pretty well for isolated unit tests.


package.json
```diff
@@ -7,7 +7,7 @@
     "build": "webpack -p",
     "dev": "webpack-dev-server --mode development",
     "lint": "eslint --ext .js src",
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "test": "tape ./test/unit/*.js"
```
Hamms: could we instead do something like

```suggestion
    "test": "npm run lint && npm run test:unit"
    "test:unit": "tape ./test/unit/*.js"
```

Issue #16, opened 2018-10-17
Reaching the end of the song should stop execution


ryansloan:

epeach:
@ryansloan I think it makes sense go back to the initial white screen here and change the 'Reset' button to 'Run' - which basically mimics a 'Reset' click. Thoughts?

epeach:
https://github.com/code-dot-org/code-dot-org/pull/25516

ryansloan:
> @ryansloan I think it makes sense go back to the initial white screen here and change the 'Reset' button to 'Run' - which basically mimics a 'Reset' click. Thoughts?

@epeach Sorry, just saw this. That sounds good to me!

PR #23, opened 2018-10-19
Add dance reset and toggle run button when song is ended

epeach:
Purpose: Halt dance when song is done

* Adds the function to toggleRunButton to the Dance Party constructor
* When the song ends, resets the dance and toggles the run button

Works with: https://github.com/code-dot-org/code-dot-org/pull/25516

### DIFF
--- a/src/p5.dance.js
+++ b/src/p5.dance.js
@@ -30,7 +30,8 @@ module.exports = class DanceParty {
     getSelectedSong,
     onPuzzleComplete,
     playSound,
-    recordReplayLog
+    recordReplayLog,
+    toggleRunButton
   }) {
     /**
      * Patch p5 tint to use fast compositing (see https://github.com/code-dot-org/p5_play/pull/42).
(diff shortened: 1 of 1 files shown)


src/p5.dance.js
```diff
@@ -178,7 +180,11 @@ module.exports = class DanceParty {
     if (this.recordReplayLog_) {
       replayLog.reset();
     }
-    this.playSound_({url: METADATA[this.getSelectedSong_()].file, callback: () => {this.songStartTime_ = new Date()}});
+    this.playSound_({
+      url: METADATA[this.getSelectedSong_()].file,
+      callback: () => {this.songStartTime_ = new Date()},
+      onEnded: () => {this.reset(); this.toggleRunButton_();}
+    });
```
joshlory: It makes sense to call `this.reset()` here, but I'm not sure we want to rely on this class to notify the caller that the song completed (seems like dependencies are going the wrong way).

Is there a way to do this with a pub/sub pattern: attaching multiple listeners to the `playSound` API, so the caller can get notified directly when the song ends?

epeach:
Closes https://github.com/code-dot-org/dance-party/issues/16

epeach:
Agreed - once the new constructor set-up is stabilized and merged, I'm going to loop back to this PR and update to match.

Issue #17, opened 2018-10-17
"Strobe" foreground effect makes dancers invisible


ryansloan:
The intended behavior was to hide the dancers when it's "Dark" and show them when it's "light", but we see no dancers at all in either state.

epeach:
https://github.com/code-dot-org/dance-party/pull/25

PR #25, opened 2018-10-19
Fixes strobe flash character disappearing

epeach:
Problem: Draws a black background over the dancers when it's 'dark' and draws a white background over the dancers when there is a 'flash'. White background covers the characters.

Solution: Only draw a background when it's 'dark'.


joshlory:
Fixes issue https://github.com/code-dot-org/dance-party/issues/17.

Issue #15, opened 2018-10-17
"(previous)" option does not work in "do move forever" block


ryansloan:

epeach:
https://github.com/code-dot-org/dance-party/pull/28

PR #28, opened 2018-10-22
Handle javascript modulo loop manually

epeach:
The prev dance was getting stuck in a loop due to Javascript using `negative % positive = negative`:
```
//when sprite.current_move = 0 and ANIMATIONS[sprite.style].length - 1 = 11
move = 1 + ((sprite.current_move - 1) % (ANIMATIONS[sprite.style].length - 1));
// move = 1 + (-1 % 11);
// move = 0;
sprite.current_move = move;
```

Due to the way Javascript handles the negative modulus case, the decrementing loop using modulus wasn't working as expected and the character did the 0th dance move infinitely. The edit handles the loop by catching the 'first' move and manually sets it back to the 'last' move.

### DIFF
--- a/src/p5.dance.js
+++ b/src/p5.dance.js
@@ -152,13 +152,21 @@ module.exports = class DanceParty {
           // it to load the image as a Image (instead of a p5.Image), which avoids
           // a canvas creation. This makes it possible to run on mobile Safari in
           // iOS 12 with canvas memory limits.
-          ANIMATIONS[this_sprite][moveIndex] = {
-            spritesheet: this.p5_.loadSpriteSheet(`${baseUrl}.png`, jsonData.frames, true),
-            mirror,
-          };
+          this.setAnimationSpriteSheet(this_sprite, moveIndex,
+            this.p5_.loadSpriteSheet(`${baseUrl}.png`, jsonData.frames, true), mirror)
         });
       });
     });
+  }
+
+  setAnimationSpriteSheet(sprite, moveIndex, spritesheet, mirror){
+    if(!ANIMATIONS[sprite]){
+      ANIMATIONS[sprite] = [];
+    }
+    ANIMATIONS[sprite][moveIndex] = {
+      spritesheet: spritesheet,
+      mirror,
+    };
   }
 
   setup() {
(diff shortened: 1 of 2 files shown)


joshlory:
Worth adding a unit regression test?

src/p5.dance.js
```diff
@@ -307,7 +307,12 @@ module.exports = class DanceParty {
     if (move === "next") {
       move = 1 + ((sprite.current_move + 1) % (ANIMATIONS[sprite.style].length - 1));
     } else if (move === "prev") {
-      move = 1 + ((sprite.current_move - 1) % (ANIMATIONS[sprite.style].length - 1));
+      //Javascript doesn't handle negative modulos as expected, so manually resetting the loop
+      if (sprite.current_move === 0) {
+        move = (ANIMATIONS[sprite.style].length - 1);
+      } else {
+        move = ((sprite.current_move - 1) % (ANIMATIONS[sprite.style].length));
```
islemaster: I don't think the modulus does anything for you at this point; I'd do one of the following to simplify:
```js
// No modulus at all
move = sprite.current_move - 1
if (move < 0) {
  move = ANIMATIONS[sprite.style].length - 1
}
// or always add the denominator to the numerator so it's never negative.
const animationCount = ANIMATIONS[sprite.style].length;
move = (animationCount + sprite.current_move - 1) % animationCount;
```

src/p5.dance.js
```diff
@@ -152,15 +152,23 @@ module.exports = class DanceParty {
...
-            mirror,
-          };
+          this.setAnimationSpriteSheet(this_sprite, moveIndex,
+            this.p5_.loadSpriteSheet(`${baseUrl}.png`, jsonData.frames, true), mirror)
         });
       });
     });
   }
 
+  setAnimationSpriteSheet(sprite, moveIndex, spritesheet, mirror){
```
epeach: I split this out into its own function so we can mock this data in unit tests.

epeach:
@joshlory and @islemaster PTAL, please!

epeach:
Closes https://github.com/code-dot-org/dance-party/issues/15

PR #30, opened 2018-10-23
Remove hard coded songs

epeach:
Before we were manually loading the development songs. Now, like in DanceVisualizationColumn, we load from the manifest.

This should also make it easier to test code using the 'testManifest.json' so we are working in an environment level, rather than at the song level.


src/p5.dance.js
```diff
@@ -685,13 +684,18 @@ module.exports = class DanceParty {
     this.setMetadata_(id, await response.json());
   }
 
-  loadDevelopmentSongs_(callback) {
-    let ids = ['macklemore90', 'hammer', 'peas'];
+  loadSongs_(callback) {
+    fetch(`/api/v1/sound-library/hoc_song_meta/songManifest.json`)
```
joshlory: Is the eventual plan to hit a different URL in test levels vs. the shipped tutorial?  Wonder if this path should be configurable (or better yet, loaded before `new DanceLab()` is called, and the manifest data passed in to the constructor instead of the manifest path).

epeach: The eventual plan was to hit different urls depending on test vs. shipped levels. This currently unblocks the work of adding/testing new songs in the manifest.

epeach:
@joshlory With the newest DanceParty API, these changes might be useful again? PTAL, thanks!

PR #31, opened 2018-10-23
Add pre- and post-version scripts

Hamms:
Also fix tests


joshlory:
> Also fix tests

Which tests are failing?  Current `master` branch build appears to be green.

Hamms:
The sanity test was failing for me because of the constructor signature change

Hamms:
Ahh, y'know, I think I was just trying to run with the wrong version of node.

Master is also passing for me; this change doesn't _fix_ the test, but it does make it more correct

Issue #35, opened 2018-10-23
next dance moves by two dances at a time instead of one-by-one.


ryansloan:
Should step through one-by-one

epeach:
https://github.com/code-dot-org/dance-party/pull/36

PR #36, opened 2018-10-23
Update Prev and Next dance move behavior

epeach:
After clarifying expected behavior for the 'prev' and 'next' dance moves:
*Next should increment by two dances.
*Prev should not use the 'rest' dance move in its cycle of dances.


epeach:
Closes https://github.com/code-dot-org/dance-party/issues/35

PR #37, opened 2018-10-24
remove metadataLoaded(), use p5.loadJSON to loadSongMetadata

cpirich:
* While modifying the share experience to autoplay, I decided to fix the load path such that it is obvious that dance-party is ready once we exit the p5 `preload()` phase. We can avoid the need for a `metadataLoaded()` and/or a new callback to indicate when metadata loading is complete if we just use `p5.loadJSON()` to load the song metadata.
* The check in `addCues()` to make sure the metadata is loaded is not needed, since `addCues()` is called within the `setup()` phase (after the `preload()` phase is complete)

### DIFF
--- a/src/p5.dance.js
+++ b/src/p5.dance.js
@@ -106,7 +106,6 @@ module.exports = class DanceParty {
     ];
 
     this.songStartTime_ = 0;
-    this.metadataLoaded_ = false;
   }
 
   pass() {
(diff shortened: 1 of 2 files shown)


joshlory:
I think this'll have conflicts with PR https://github.com/code-dot-org/dance-party/pull/34.  Did you see [the thread in Slack](https://codedotorg.slack.com/archives/CC2QHV7SR/p1540322474000100) with the planned API changes?

PR #42, opened 2018-10-24
New dance api phase1

cpirich:
API Changes:
* `play(songData)` - this is what the host will call when it is ready to start playing
* `addCues()` no longer necessary to generate `peaksData`
* No more `loadSongMetadata_()`, `getSelectedSong`, `loadDevelopmentSongs_()`, etc.
* No more `metadataLoaded()`


src/p5.dance.js
```diff
@@ -80,7 +77,6 @@ module.exports = class DanceParty {
     this.world.fg_effect = null;
 
     //TODO - need to reset on run or load new song
```
epeach: Nit: we can remove this along with peaksData

PR #44, opened 2018-10-24
New Background effects

ryansloan:
Adding some effects requested by LT and others, as well as unifying some color palette choices per Mark's feedback


src/Effects.js
```diff
@@ -191,5 +191,72 @@ module.exports = class Effects {
...
+        }
+      }
+    };
+    this.spiral = {
+      swirl: null,
+      angle: 0,
+      color: null,
+      init: function () {
+        //I should do this somewhere else, right?
+        this.swirl = p5.loadImage("https://curriculum.code.org/images/spiral.svg");
```
joshlory: I wonder if this is a small enough asset that we should just include it inline?

Sample: https://editor.p5js.org/joshlory/sketches/HkH6IjRjm

src/Effects.js
```diff
@@ -191,5 +191,72 @@ module.exports = class Effects {
...
+        p5.strokeWeight(0);
+      },
+      update: function () {
+        //TODO: add some music-driven change? Right now it just grows continuously.
+      },
+      draw: function () {
+        if (this.splats.length<1) {
+          this.init();
+        }
+        p5.svg();
```
joshlory: Hmm, I'm getting `p5.svg is not a function`.  Different version of p5.js?

joshlory:
```js
nativeAPI.setBackgroundEffect('spiral');
```

image

Cool! 🎉

ryansloan:
> If mark has color palette ideas it would be nice to just add an array that could be used by all of the effects. Maybe even even add an “effect theme” block so you could programmatically select from a set of palettes...

@mrjoshida I did get a palette from Mark that was based on contrast with the character palette, but after implementing it we realized it didn't look as good as the existing color generation scheme you were using. He suggested we continue to generate colors following the same algorithm, and we can do palette tweaking later.

I do think it would be good to refactor some of the Effects class to centralize color generation/picking though

Issue #43, opened 2018-10-24
"after N measures" event hat positioning impacts code sequence


ryansloan:
Create this program
<img width="647" alt="screen shot 2018-10-24 at 4 04 25 pm" src="https://user-images.githubusercontent.com/1669271/47466543-858db600-d7a6-11e8-897c-19e165dd1f95.png">

Notice that the program runs "correctly" (characters floss before they clap)

Now move the event code so that the "after 2 measures" is above the "after 1 measure", like so:
<img width="675" alt="screen shot 2018-10-24 at 4 04 54 pm" src="https://user-images.githubusercontent.com/1669271/47466556-9c340d00-d7a6-11e8-8775-ffce25a2e5a0.png">

Notice that the characters now start flossing after 2 measures (instead of 1) and never clap.

PR #49, opened 2018-10-26
Bug Fix: Block order doesn't change dance

epeach:
Addresses: https://github.com/code-dot-org/dance-party/issues/43

* Sorts cues (with test)

* After discussion with JoshL, adds p5.dance.interpreted.js to dance-party

### DIFF
--- a/src/p5.dance.js
+++ b/src/p5.dance.js
@@ -112,11 +112,9 @@ module.exports = class DanceParty {
   addCues(timestamps) {
     // Sort and de-duplicate
     const numSort = (a,b) => a - b;
-    let measures = timestamps.measures.sort(numSort);
-    let seconds = timestamps.seconds.sort(numSort);
 
-    this.world.cues.measures = measures.filter((measure, index) => measure != measures[index - 1]);
-    this.world.cues.seconds = seconds.filter((time, index) => time != seconds[index - 1]);
+    this.world.cues.measures = timestamps.measures.sort(numSort);
+    this.world.cues.seconds = timestamps.seconds.sort(numSort);
   }
 
   reset() {
--- a/src/p5.dance.js
+++ b/src/p5.dance.js
@@ -112,7 +112,13 @@ module.exports = class DanceParty {
   }
 
   addCues(timestamps) {
-    this.world.cues = timestamps;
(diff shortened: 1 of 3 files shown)


src/p5.dance.js
```diff
@@ -111,7 +110,11 @@ module.exports = class DanceParty {
   }
 
   addCues(timestamps) {
-    this.world.cues = timestamps;
+    // Sort and de-duplicate
+    const numSort = (a,b) => a - b;
```
joshlory: Also need to de-duplicate?

test/unit/cuesTest.js
```diff
@@ -0,0 +1,34 @@
+const test = require('tape');
+const DanceParty = require('../../src/p5.dance');
+
+const createDanceAPI = () => {
```
joshlory: Nit: we're starting to duplicate this a number of places, I wonder if its worth extracting to a test/helpers util?

(Non-blocking, we could address this in a follow up PR.)

test/unit/cuesTest.js
```diff
@@ -0,0 +1,34 @@
...
+  nativeAPI.play({
+    bpm: 120,
+  });
+
+  //Initial Value
+  t.deepEqual(nativeAPI.world.cues.measures, []);
+  t.deepEqual(nativeAPI.world.cues.seconds, []);
+
+  let cues = {measures: [0, 1, 4, 2, 0, 8, 3, 3, 3], seconds: [0, 1, 4, 2, 0, 8, 3, 3, 3]};
+  let sortedCues = [0, 0, 1, 2, 3, 3, 3, 4, 8];
```
joshlory: Should be `[0, 1, 2, 3, 4, 8];` if we de-duplicate?

src/p5.dance.js
epeach: Whoops! Leftover comment from when I thought we'd need to de-duplicate. The while loop in updateEvents means that duplicates are already handled as long as they are sorted. I'll update that comment.

joshlory: Don't we want to de-duplicate?  I guess it's fine for the `while` loop to take care of it... but sort of prefer to sanitize the data.

PR #50, opened 2018-10-26
Update p5play add events and key support

cpirich:
* Cleaned up how we get our version of p5 (it now comes in a sub-dependency of p5.play)
* Add 2 new API methods: `onKeyDown(keyCode)` and `onKeyUp(keyCode)` - these are used to route key events from the host down into dance/p5 (this is used when the host wants soft buttons)
* Add a new `showMeasureLabel` option that defaults to `true` - we are likely to make this `false` in cdo's share mode
* Add the ability for the host to provide an `onHandleEvents` function that will be called each frame as long as there are new events (the host doesn't need to know about `currentFrameEvents` - they will be passed as a param to `onHandleEvents()`)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@code-dot-org/dance-party",
-  "version": "0.0.7",
+  "version": "0.0.8",
   "description": "",
   "main": "dist/main.js",
   "scripts": {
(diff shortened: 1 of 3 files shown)


src/loadP5.js
```diff
@@ -1,46 +1,14 @@
...
-    tmpCtx.fillStyle = 'hsl(' + this._pInst.hue(this._tint) + ', 100%, 50%)';
-    tmpCtx.fillRect(0, 0, this._tintCanvas.width, this._tintCanvas.height);
-    tmpCtx.globalCompositeOperation = 'destination-atop';
-    tmpCtx.drawImage(img.canvas, 0, 0, this._tintCanvas.width, this._tintCanvas.height);
-    tmpCtx.globalCompositeOperation = 'multiply';
-    tmpCtx.drawImage(img.canvas, 0, 0, this._tintCanvas.width, this._tintCanvas.height);
-    return this._tintCanvas;
-  };
-}
+const P5 = require('@code-dot-org/p5');
```
joshlory: So much cleaner!

package.json
```diff
@@ -21,13 +21,12 @@
   "author": "",
   "license": "Apache-2.0",
   "devDependencies": {
-    "@code-dot-org/p5.play": "1.3.4-cdo",
+    "@code-dot-org/p5.play": "1.3.5-cdo",
```
joshlory: Need to add a `@code-dot-org/p5` dev dependency?

cpirich: I had one, but removed it because it was implied as a sub-dependency of p5.play. If I put one here, I'll need to keep it in sync with the version that p5.play is pulling in.

PR #54, opened 2018-10-27
Add integration tests for more HoC levels

joshlory:
Following the pattern set up in PR https://github.com/code-dot-org/dance-party/pull/53.

### DIFF
--- a/levels/hourOfCode.js
+++ b/levels/hourOfCode.js
@@ -119,4 +119,41 @@ module.exports = {
       }
     `,
   },
+  hoc09: {
+    solution: `
+      var left_shark;
+      var right_pineapple;
+
+      whenSetup(function () {
+        setBackgroundEffect("rainbow");
+        left_shark = makeNewDanceSprite("MOOSE", left_shark, {x: 100, y: 200});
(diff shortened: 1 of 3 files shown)


levels/hourOfCode.js
```diff
@@ -53,4 +53,70 @@ module.exports = {
...
+    validationCode: `
+      if (nativeAPI.getTime("measures") > 7) {
+        if (World.bg_effect == null) {
+          nativeAPI.fail("You need to add a background effect.");
+        }
+      }
+      if (nativeAPI.getTime("measures") > 10) {
+        nativeAPI.pass();
+      }
+    `,
```
joshlory: I'm pulling the solution from the "Show Code" button in the workspace.  The validation code is copied from `appOptions.level.validationCode`.

joshlory:
Good question!  In the Craft repo the test collateral levels are also totally disconnected from the live-site levels.  My idea behind these tests is to give a representative set of actions to exercise different code paths — not necessarily verify the live-site progression is working.  For that we should probably rely on tests that run when the level is saved (faster feedback to level builders) or realtime pass rates for users on production.

If we updated a live-site level, we might want to keep the old integration test around in addition to the new one (depending on coverage).

In practices, these levels shouldn't change much going forward, so they're effectively locked to the production levels.

PR #67, opened 2018-10-30
add short burst moves, test hooks, and babel-core dev dependency

cpirich:
* We now have "short burst" moves uploaded for all characters except pineapple and alien (which are using placeholder cat animations with the expected filenames). Adding these to `MOVE_NAMES` so we can expose them with blocks. Testing has verified that we can load all of these moves within the RAM limits on a "low end" device (verified on iPad Mini 2 with iOS 10.1.1 and on iPhone 5s with iOS 12.0.1).
* Added test hooks to make it easier to detect the number of sprites and the song playing from within UI tests: `window.__mostRecentDancePartySongUrl` and `window.__mostRecentDancePartySpriteCount`
* Explicitly added `babel-core` as a dev dependency. This makes our project compatible with `yarn` as well as `npm`

### DIFF
--- a/src/p5.dance.js
+++ b/src/p5.dance.js
@@ -85,6 +85,12 @@ module.exports = class DanceParty {
       {name: "Roll", mirror: true},
       {name: "ThisOrThat", mirror: false},
       {name: "Thriller", mirror: true},
+      {name: "XArmsSide", mirror: true},
+      {name: "XArmsUp", mirror: true},
+      {name: "XJump", mirror: true},
+      {name: "XClapSide", mirror: true},
+      {name: "XHeadHips", mirror: true},
+      {name: "XHighKick", mirror: true},
     ];
 
     if (spriteConfig) {
@@ -199,6 +205,9 @@ module.exports = class DanceParty {
     this.analysisPosition_ = 0;
     this.playSound_({url: this.songMetadata_.file, callback: () => {this.songStartTime_ = new Date()}});
     this.p5_.loop();
+
+    /** Expose for testing **/
+    window.__mostRecentDancePartySongUrl = this.songMetadata_.file;
   }
 
   setBackground(color) {
@@ -798,5 +807,8 @@ module.exports = class DanceParty {
     if (this.currentFrameEvents.any && this.onHandleEvents) {
       this.onHandleEvents(this.currentFrameEvents);
     }
+
+    /** Expose for testing **/
+    window.__mostRecentDancePartySpriteCount = this.p5_.allSprites.length;
   }
 };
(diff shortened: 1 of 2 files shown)


cpirich:
@ryansloan FYI

ryansloan:
Huzzah! I'll own subsequent block updates on LB

src/p5.dance.js
```diff
@@ -199,6 +205,9 @@ module.exports = class DanceParty {
     this.analysisPosition_ = 0;
     this.playSound_({url: this.songMetadata_.file, callback: () => {this.songStartTime_ = new Date()}});
     this.p5_.loop();
+
+    /** Expose for testing **/
+    window.__mostRecentDancePartySongUrl = this.songMetadata_.file;
```
joshlory: Should we expose this on the code-dot-org side?

src/p5.dance.js
```diff
@@ -798,5 +807,8 @@ module.exports = class DanceParty {
     if (this.currentFrameEvents.any && this.onHandleEvents) {
       this.onHandleEvents(this.currentFrameEvents);
     }
+
+    /** Expose for testing **/
+    window.__mostRecentDancePartySpriteCount = this.p5_.allSprites.length;
```
joshlory: I'd rather expose `nativeAPI` in Dance.js — it seems like a concern thats tied to our code-dot-org UI tests, not so much this repo.

joshlory:
Changes (1) and (3) could prob be their own PR (in the future, no need to change for this PR), to keep everything separable!

src/p5.dance.js
cpirich: @joshlory 
(1) I don't want to move objects to `window` as it creates object leaks (making it hard to debug)
(2) I want to make it possible to use this info for dance-party unit tests OR for cdo UI tests

joshlory: Re: (1) can we only do so for test environments?

src/p5.dance.js
cpirich: It is certainly possible to only expose this only on the cdo side if we don't want access to it for unit tests on this side. I'm not sure that buys us a whole lot.

src/p5.dance.js
cpirich: maybe? it doesn't appear that any other cdo repo code does this (looking at `window.__TestInterface` as the primary example). We'd need to route environment info in here through a constructor param.

cpirich: Also, I dismissed the `registerValidation()` approach from the cdo side because it would interfere with the real validation code. The dance API only allows one `validationCallback` to be registered at a time. Without changing that, we'd have to then wrap the level's validation code with our own "test hook" that calls the level's validation code, which seems ugly.

joshlory: The DanceParty class itself doesn't need to know if it should be in test mode, I think we should expose the whole `nativeAPI` object like we do for Game Lab.

PR #79, opened 2018-10-30
Add test hook interface

cpirich:
* Add a new test interface via `DanceParty.getTestInterface()` that will allow test code in the `code-dot-org` repo to access some of the internal state of `DanceParty` without poking into private properties and methods that are subject to change.
* The interface currently consists of `getSprites()`, `getSongUrl()`, and `getSongStartedTime()` - and can be extended over time.

### DIFF
--- a/src/p5.dance.js
+++ b/src/p5.dance.js
@@ -124,6 +124,14 @@ module.exports = class DanceParty {
     }
   }
 
+  getTestInterface() {
+    return {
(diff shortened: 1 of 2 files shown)


joshlory:
Mind merging master into this branch, to reduce the diff?

Or better yet, rebase this branch against master so it only contains commit b13573a4614a6497030070b88c14c186340c7d6e.

Issue #74, opened 2018-10-30
Support for more input keys


ryansloan:
Ordered: up, down, left, right, space, enter, a-z, 0-9

ryansloan:
Fixed with #80 !

ryansloan:
Updated in Levelbuilder

PR #80, opened 2018-10-30
check more keys

Bjvanminnen:
https://github.com/code-dot-org/dance-party/issues/74

This PR makes it so that our when pressed key has many more options:
image

This will also require an LB change, with the new config below. I validated that this worked for me locally.

```json
{
  "func": "whenKey",
  "blockText": "when {KEY} pressed",
  "args": [
    {
      "name": "KEY",
      "options": [
        [
          "up",
          "\"up\""
        ],
        [
          "left",
          "\"left\""
        ],
        [
          "down",
          "\"down\""
        ],
        [
          "right",
          "\"right\""
        ],
        [
          "space",
          "\"space\""
        ],
        [
          "enter",
          "\"enter\""
        ],
        [
          "a",
          "\"a\""
        ],
        [
          "b",
          "\"b\""
        ],
        [
          "c",
          "\"c\""
        ],
        [
          "d",
          "\"d\""
        ],
        [
          "e",
          "\"e\""
        ],
        [
          "f",
          "\"f\""
        ],
        [
          "g",
          "\"g\""
        ],
        [
          "h",
          "\"h\""
        ],
        [
          "i",
          "\"i\""
        ],
        [
          "j",
          "\"j\""
        ],
        [
          "k",
          "\"k\""
        ],
        [
          "l",
          "\"l\""
        ],
        [
          "m",
          "\"m\""
        ],
        [
          "n",
          "\"n\""
        ],
        [
          "o",
          "\"o\""
        ],
        [
          "p",
          "\"p\""
        ],
        [
          "q",
          "\"q\""
        ],
        [
          "r",
          "\"r\""
        ],
        [
          "s",
          "\"s\""
        ],
        [
          "t",
          "\"t\""
        ],
        [
          "u",
          "\"u\""
        ],
        [
          "v",
          "\"v\""
        ],
        [
          "w",
          "\"w\""
        ],
        [
          "x",
          "\"x\""
        ],
        [
          "y",
          "\"y\""
        ],
        [
          "z",
          "\"z\""
        ],
        [
          "0",
          "\"0\""
        ],
        [
          "1",
          "\"1\""
        ],
        [
          "2",
          "\"2\""
        ],
        [
          "3",
          "\"3\""
        ],
        [
          "4",
          "\"4\""
        ],
        [
          "5",
          "\"5\""
        ],
        [
          "6",
          "\"6\""
        ],
        [
          "7",
          "\"7\""
        ],
        [
          "8",
          "\"8\""
        ],
        [
          "9",
          "\"9\""
        ]
      ]
    }
  ],
  "eventBlock": true,
  "color": [
    140,
    1,
    0.74
  ]
}
```

### DIFF
--- a/test/unit/whenKeyTest.js
+++ b/test/unit/whenKeyTest.js
@@ -6,13 +6,13 @@ test('make sure we are checking all relevant keys', async t => {
   const p5 = nativeAPI.p5_;
 
   // simulate key presses
-  p5._onkeydown({which: p5.KEY.LEFT_ARROW});
-  p5._onkeydown({which: p5.KEY.SPACE});
-  p5._onkeydown({which: p5.KEY.ENTER});
-  p5._onkeydown({which: p5.KEY.A});
-  p5._onkeydown({which: p5.KEY.Z});
-  p5._onkeydown({which: p5.KEY['0']});
-  p5._onkeydown({which: p5.KEY['9']});
+  nativeAPI.onKeyDown(p5.KEY.LEFT_ARROW);
+  nativeAPI.onKeyDown(p5.KEY.SPACE);
+  nativeAPI.onKeyDown(p5.KEY.ENTER);
+  nativeAPI.onKeyDown(p5.KEY.A);
+  nativeAPI.onKeyDown(p5.KEY.Z);
+  nativeAPI.onKeyDown(p5.KEY['0']);
+  nativeAPI.onKeyDown(p5.KEY['9']);
   p5.readPresses();
 
   nativeAPI.updateEvents_();
--- a/test/unit/whenKeyTest.js
+++ b/test/unit/whenKeyTest.js
@@ -1,0 +1,31 @@
+const test = require('tape');
+const helpers = require('../helpers/createDanceAPI');
+
+test('make sure we are checking all relevant keys', async t => {
+  const nativeAPI = await helpers.createDanceAPI();
+  const p5 = nativeAPI.p5_;
+
+  // simulate key presses
+  p5._onkeydown({which: p5.KEY.LEFT_ARROW});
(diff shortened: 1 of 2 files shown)


test/unit/whenKeyTest.js
```diff
@@ -0,0 +1,31 @@
...
+
+  // simulate key presses
+  p5._onkeydown({which: p5.KEY.LEFT_ARROW});
+  p5._onkeydown({which: p5.KEY.SPACE});
+  p5._onkeydown({which: p5.KEY.ENTER});
+  p5._onkeydown({which: p5.KEY.A});
+  p5._onkeydown({which: p5.KEY.Z});
+  p5._onkeydown({which: p5.KEY['0']});
+  p5._onkeydown({which: p5.KEY['9']});
+  p5.readPresses();
```
Bjvanminnen: You could make the argument that this test is overly brittle, as we're using p5 internals to simulate our presses, but I think this is likely still better than nothing.

joshlory: I wonder if it's better to call `nativeAPI.onKeyDown` here?

Bjvanminnen: Agree entirely :)

cpirich:
If students write programs that use these other keys, they won’t work well when shared to phones. I just got the soft buttons restored for the Share scenario, but is only the arrows.

Welcome back, Brent! Just heard from Miranda this afternoon that you’re helping out!!

ryansloan:
Just chiming in here in response to @cpirich's comments - we are only requiring the use of arrows in the tutorial (which we have the soft keys for). But we want to open up the capability for students to use all the keys on their devices. We do this in the curriculum for apps like Playlab as well.

PR #82, opened 2018-10-31
Artist / Title background effect

mrjoshida:
We didn't get lyrics 😞 but I still want a text focused background effect. I've added "artist" and "title" fields to all existing metadata files to allow for this.

_Note_: this is based on a background effect submitted by a CSD student!

### DIFF
--- a/metadata/hammer.json
+++ b/metadata/hammer.json
@@ -1,5 +1,7 @@
 {
   "file": "https://curriculum.code.org/media/uploads/touch.mp3",
+  "artist": "MC Hammer",
+  "title": "U Can't Touch This",
   "bpm": "133",
   "delay": "2.32",
   "analysis": [
--- a/levels/effectsLevels.js
+++ b/levels/effectsLevels.js
@@ -1,0 +1,38 @@
+module.exports = {
+  changeBackgroundAtTimestamp: {
+    solutions:
+      `
+      var lead_dancer;
+
+      whenSetup(function () {
+        setBackgroundEffect("color_cycle");
+      });
+      
+      atTimestamp(4, "measures", function () {
+        setBackgroundEffect("text");
+      });
+    `,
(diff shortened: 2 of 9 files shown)


mrjoshida:
Added a new foreground effect inspired by Jael's promo gif
color lights

joshlory:
@Hamms putting this on your radar because it has video generation implications.
 - Will students be sad if they can't see the light (foreground) effects in their exported video?
 - Any concerns with rendering text in the video?

metadata/hammer.json
```diff
@@ -1,5 +1,7 @@
 {
   "file": "https://curriculum.code.org/media/uploads/touch.mp3",
+  "artist": "MC Hammer",
```
epeach: We have deprecated the use of metadata files in dance-party (other than our test songs - synthesize and jazzy_beats) so we don't need to update these files and they should, at some point, be removed from the repo.

Hamms:
> Will students be sad if they can't see the light (foreground) effects in their exported video?

I'm sure they will be; as I mentioned in our chat earlier, I'm going to try to get the foreground effects in, but am not planning to spend time right now trying to fix foreground effects that don't work right off the bat

> Any concerns with rendering text in the video?

Hmmm, possibly. I know rendering emojis don't work in the serverside p5, but regular text should be fine. I'll test in the text couple of days

metadata/hammer.json
mrjoshida: Should I pull the extras out as part of the PR?

mrjoshida:
One more bg effect - replaced "rainbow" with an actual rainbow, renamed the old rainbow "color_cycle"
rainbow

metadata/hammer.json
epeach: We'll want to at some point, but that can come in a follow-up PR at some point.

Issue #73, opened 2018-10-30
Additional character layouts


ryansloan:
**In all layouts, Characters will be resized to fit in the view, down to .3 scale**
Sorry for the humongous pictures, I didn't realize when I snapped them and uploaded them that my phone went all out. 

# "Outside Border"
This will create a square border of characters around the screen, with the middle open. It will fill in the following order:
![First Four sprites for outside border](https://user-images.githubusercontent.com/1669271/47760483-b9b71a00-dc71-11e8-8b48-c97034ce1305.png)

After they have more than 8 characters, it will fill in clockwise in the middle of each gap:
![9+ sprites for outside border](https://user-images.githubusercontent.com/1669271/47760503-cfc4da80-dc71-11e8-9c1a-1242f9bf60f6.png)

# "Filled Square"
The filled square layout will grow out from the center. It will fill in starting from the top right, and will proceed right all the way to the last column, then down to the next row. The target square will be MxM where `M = Math.ceil(Math.sqrt(sprites.length))`. Example growth:

![Growth for Filled Square](https://user-images.githubusercontent.com/1669271/47760530-ebc87c00-dc71-11e8-9569-d4e7bad9fa8d.png)

# "Circle" 
The Circle layout already exists, but should be updated to do the following:

* Resize the sprites to fit 
* Rotate them so that they are all angled along the circle's radius at that point. (The circles in the diagram below are the "heads")
* Order such that z-index of sprites grows with their Y position (ex: a sprite that is on the bottom of the canvas at [200, 380] should be on a layered above a sprite that is higher on the canvas at [200,300]
* Place sprites so that they are positioned in steps of 360/N degrees

# "Plus"
The plus layout will grow out from the center, ordering each new character from top->bottom->left->right->top...

![Plus Layout growth](https://user-images.githubusercontent.com/1669271/47761084-5e3a5b80-dc74-11e8-9ce3-9418a64be56b.jpg)

# "X"
The X layout will grow from the center like the plus, but with a few key differences:
* Characters will be positioned 45 degree angles from the origin instead of 90 degree angles
* Characters will be rotated to match their angle of positioning (the circles in the diagram below are the "heads")

![X Layout Growth](https://user-images.githubusercontent.com/1669271/47761317-254eb680-dc75-11e8-9b10-e648b36e60f5.JPG)

# original sketches
The designs above differ slightly from these sketches, because the designs were scaled to allow for an arbitrary number of sprites.

[...]

PR #86, opened 2018-10-31
Layouts

Bjvanminnen:
https://github.com/code-dot-org/dance-party/issues/73

This PR makes changes to a couple of existing layouts and also adds some new ones.

After laying sprites out, we also update the z-index (depth) based on the sprite position.

For most of the sprites we set them to size 30 before laying them out. In the case of the circle, we might then resize them according to number of sprites.

Each layout has been tested for a variety of sprite counts.

We'll also want to make the following LB changes
- Update layout block
- Update group create block
- No longer manually size in the tutorial (right now we have a block that sets to size 50 after group creation).

Here are the appropriate layout options
```
{
      "name": "LAYOUT",
      "options": [
        [
          "circle",
          "\"circle\""
        ],
        [
          "border",
          "\"border\""
        ],
        [
          "grid",
          "\"grid\""
        ],
        [
          "inner square",
          "\"inner\""
        ],
        [
          "row",
          "\"row\""
        ],
        [
          "column",
          "\"column\""
        ],
        [
          "plus",
          "\"plus\""
        ],
        [
          "X",
          "\"x\""
        ],
        [
          "random",
          "\"random\""
        ]
      ]
    }
```

Some minimal discussion of each layout follows, along with numerous screenshots

## circle
Sprites are evenly spaced out in a circle, aligned with their feet towards the center. We vary the size of the circle according to the number of sprites in it. In addition, we vary the sprite site.

count 3
image

count 10
image

count 100
image

## border
We start by placing sprites in the corners. We then divide the remaining sprites into four groups as evenly as possible, and distribute them along the various edges

count 5
image

count 10
image

count 100
image

## grid
Made some changes here, the biggest one being that if we have a non-full row, instead of centering it, we left align it, i.e. instead of
XXX
 XX
we do
XXX
XX

count 5
image

count 10
image

count 100
image

## Inner square
The square size varies some according to count, but never gets larger than half the play area

count 3
image

count 10
image

count 100
image

## plus
Order is top, bottom, left, right

count 5
image

count 10
image

count 100
image

## X
similar to plus but rotated 45 degrees. also, the sprites themselves are rotated to have the feet face inwards

count 5
image

count 10
image

count 100
image

### DIFF
--- a/src/p5.dance.js
+++ b/src/p5.dance.js
@@ -494,8 +494,9 @@ module.exports = class DanceParty {
       group.forEach((sprite, i) => {
         const row = Math.floor(i / numCols);
         const col = i % numCols;
-        sprite.x = this.p5_.lerp(minX, maxX, col / (numCols - 1));
-        sprite.y = this.p5_.lerp(minY, maxY, row / (numRows - 1));
+        // || 0 so that we recover from div 0
+        sprite.x = this.p5_.lerp(minX, maxX, col / (numCols - 1) || 0);
+        sprite.y = this.p5_.lerp(minY, maxY, row / (numRows - 1) || 0);
       });
     } else if (format === "inner") {
       const pct = this.p5_.constrain(count / 10, 0, 1);
--- a/src/p5.dance.js
+++ b/src/p5.dance.js
@@ -429,25 +429,60 @@ module.exports = class DanceParty {
     const maxX = 400 - minX;
     const minY = 35;
     const maxY = 400 - 40;
+    const radiansToDegrees = 180 / Math.PI;
+    const maxCircleRadius = 165;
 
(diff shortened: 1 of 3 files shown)


src/p5.dance.js
```diff
@@ -413,62 +413,180 @@ module.exports = class DanceParty {
...
       });
+    } else {
+      throw new Error('Unexpected format: ' + format);
     }
+
+    // we want sprites that are lower in the canvas to show up on top of those
+    // that are higher
+    // we also add a fractional component based on x to avoid z-fighting (except
+    // in cases where we have identical x and y)
+    // TODO: add test for x portion
```
Bjvanminnen: TODO

PR #88, opened 2018-10-31
Fix setBackground

### LINKED ISSUE #77: Background color is broken
screen shot 2018-10-30 at 12 28 04 pm

madelynkasula:
Fixes #77.

There are two parts to this fix:
- If `bg_effect` isn't set when we call `draw()`, we don't call `this.world.bg_effect.draw(context)`, so background color (which is in context) doesn't get set. _Solution: Set p5 background color to `this.world.background_color` if it is set._
- If `bg_effect` is set when `setBackground(color)` was called, the background effect "wins" and the new background color isn't visible. _Solution: Clear the background effect (set `bg_effect` to "none") when setting a background color._

### DIFF
--- a/src/p5.dance.js
+++ b/src/p5.dance.js
@@ -220,6 +220,8 @@ module.exports = class DanceParty {
   }
 
   setBackground(color) {
+    // Clear background effect so it doesn't cover up background color.
+    this.world.bg_effect = this.bgEffects_.none;
     this.world.background_color = color;
   }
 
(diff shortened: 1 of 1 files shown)


joshlory:
@madelynkasula Plan for test coverage?

madelynkasula:
@joshlory do you have a suggestion? i'm unfamiliar in this area, but [`cuesTest.js`](https://github.com/code-dot-org/dance-party/blob/master/test/unit/cuesTest.js) looks like an appropriate place to add unit tests for `setBackground`

joshlory:
I think `cueTests.js` is for verifying "after 4 measure ... do something".  I'd recommend [`sanity.js`](https://github.com/code-dot-org/dance-party/blob/master/test/unit/sanity.js) or a new file specifically for unit testing of effects!  Right now I think we only have integration test coverage of effects/backgrounds, no unit tests.

Issue #87, opened 2018-10-31
When rendering short burst moves, spritesheets flicker on screen


ryansloan:
Example project: cMFiZ9Q2Nr2fDLsrzVRW0w

Screen capture: (It’s expected that there are ducks placed randomly, but the flashing strip of ducks is not expected)

![glitchy_spritesheets](https://user-images.githubusercontent.com/1669271/47824186-d4989580-dd28-11e8-910c-5fba353d9542.gif)

ryansloan:
Assigning to @cpirich because he's the one who most recently made changes related to spritesheets, not sure what caused this though.

ryansloan:
Seems like two issues here:
1) `changeMoveLR()` should not include short burst moves in its list of moves for next/previous/random
2) Short burst moves aren't working for any characters except the Cat (and Alien and Pineapple, since they become the cat when doing short moves right now)

cpirich:
#1 above is addressed by https://github.com/code-dot-org/dance-party/pull/101
#2 is related to a script issue where the spritesheets were inadvertently resized. I've uploaded new spritesheets, but didn't both fixing Ailen / Pineapple since we're still waiting for real assets there.

cpirich:
Just keep in mind that PR#101 is not yet merged, so the issue can't be "closed" quite yet...

PR #101, opened 2018-11-02
short burst and rand fixes

cpirich:
* Fixing this: https://github.com/code-dot-org/dance-party/issues/87
* We had some special behavior that needed to be cleaned up for maintenance / readability:
  * `next` / `prev` / `rand` on `changeMoveLR()` (do forever) are supposed to exclude the "resting" animation.
    * Cleaned up how we determine which animation(s) are "resting" and make it more explicit (rather than magical `1 +` and `<= 0` conditions) : we now have `restMoveCount`
    * Fixed `rand` to be consistent with `next` and `prev` (confirmed with @mrjoshida )
  * `next` / `prev` / `rand` are supposed to exclude "short burst" animations altogether.
    * In a similar way, cleaned up how we determine which animations are safe to switch between : we now have `fullLengthMoveCount`, which include "resting" move(s)
* `doMoveLR()` (do once) had some special code to jump into the middle of the animation (`FRAMES / 2`) that should not apply to short burst moves since they have only half of the frames. That is now fixed.
* Updated tests that bypass the normal init flow to ensure that `this.world.MOVE_NAMES[]` and the new `this.world.fullLengthMoveCount` and `this.world.restMoveCount` properties are set reasonably so the test will run and pass as expected

### DIFF
--- a/src/p5.dance.js
+++ b/src/p5.dance.js
@@ -24,6 +24,7 @@ const SIZE = 300;
 const ANIMATIONS = {};
 const FRAMES = 24;
 
+// NOTE: min and max are inclusive
 function randomInt(min, max) {
   return Math.floor(Math.random() * (max - min + 1)) + min;
 }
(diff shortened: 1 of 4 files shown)


epeach:
Looks good! Now that we specify the behavior for error cases like invalid move numbers and needing >2 dance moves, can we add some tests to cover those cases?

epeach:
The new tests look good - thanks! I think it might be with while to add test coverage on the doMoveLR short burst case to make sure we don't see a regression in that area.

PR #106, opened 2018-11-03
Track animation state directly on the sprite, instead of on animation

joshlory:
Prep for alternate draw methods, besides p5.play animation.

### DIFF
--- a/src/p5.dance.js
+++ b/src/p5.dance.js
@@ -264,15 +264,19 @@ module.exports = class DanceParty {
     this.sprites_by_type_[costume].add(sprite);
 
     sprite.mirroring = 1;
+    sprite.isLooping = true;
     sprite.looping_move = 0;
     sprite.looping_frame = 0;
     sprite.current_move = 0;
     sprite.previous_move = 0;
 
-    for (var i = 0; i < ANIMATIONS[costume].length; i++) {
+    for (let i = 0; ANIMATIONS[costume] && i < ANIMATIONS[costume].length; i++) {
       sprite.addAnimation("anim" + i, ANIMATIONS[costume][i].animation);
     }
-    sprite.animation.stop();
+    if (sprite.animation) {
+      sprite.animation.stop();
+    }
+    sprite.frame = 0;
     this.sprites_.add(sprite);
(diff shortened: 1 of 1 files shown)


src/p5.dance.js
```diff
@@ -287,30 +291,38 @@ module.exports = class DanceParty {
       var msPerBeat = 60 * 1000 / (this.songMetadata_.bpm * (sprite.dance_speed / 2));
       var msPerFrame = msPerBeat / FRAMES;
       while (sprite.sinceLastFrame > msPerFrame) {
+        const totalFrames = sprite.current_move > 12 ? 12 : 24;
```
islemaster: Two problems here:

1. Should be `>= 12`; our animations are zero-indexed.
2. `current_move` is used to track the last _looping_ move the player set, and to return to it after a one-time move is complete.  Unfortunately we started using it to track the looping move.  We'll want to separate these.

src/p5.dance.js
```diff
@@ -365,8 +377,8 @@ module.exports = class DanceParty {
     sprite.mirroring = dir;
     sprite.mirrorX(dir);
     sprite.changeAnimation("anim" + move);
```
islemaster: This logs an error when `?vector=true`

src/p5.dance.js
```diff
@@ -287,30 +291,38 @@ module.exports = class DanceParty {
...
             sprite.mirrorX(sprite.mirroring);
           }
         }
 
-        var currentFrame = sprite.animation.getFrame();
-        if (currentFrame === sprite.animation.getLastFrame() && !sprite.animation.looping) {
-          //changeMoveLR(sprite, sprite.current_move, sprite.mirroring);
-          sprite.changeAnimation("anim" + sprite.current_move);
-          sprite.animation.changeFrame(sprite.looping_frame % sprite.animation.images.length);
+        if (currentFrame === totalFrames && !sprite.isLooping) {
```
islemaster: This wasn't working for me - I think because we set sprite.frame `% totalFrames` above so they'll never be equal.  You could compare to totalFrames - 1 but I think that drops the last frame in a non-looping animation?

PR #110, opened 2018-11-04
SVG Dancers

islemaster:
With lots of work from @joshlory.

Behind a queryparam `?vector=true` renders from SVG instead of PNG spritesheets.

Currently:
- Downloads SVG
- During preload creates empty images at appropriate size and returns them for creating spritesheets.
- During preload renders spritesheets from SVG into previously created image objects (as base64 png datauris).
- Preload blocks until conversion to dataURLs is done.
- Pseudo-trims spritesheets (cuts off limbs in a few cases)

Could explore:
- Lazy-render spritesheets from SVG as they are requested.
- Real texture packing
- Render at lower quality at first and improve it later.
- Cache dataURIs to localStorage so cold-start cost is only paid once.

Wins:
- Total download size ~1MB, about 1/10th the size of the PNG spritesheets.
- Potential for very fast startup for small cases.

Problems:
- Up-front rendering is slow (~11s on a fast machine).  Maybe better than cold-start download of PNG spritesheets, but still uncomfortably slow.
- Still worse performance and worse memory footprint in the worst case (rendering all animations at once).


src/p5.dance.js
```diff
@@ -3,6 +3,9 @@
 const P5 = require('./loadP5');
 const Effects = require('./Effects');
 const replayLog = require('./replay');
+const SpritesheetBuilder = require('./SpritesheetBuilder');
+
+const VECTOR_MODE = window.location.search.startsWith('?vector=');
```
Bjvanminnen: Presumably this is meant to semi-imitate what we do with experiments in our main repo. Might be worth a short comment that this is a temporary hack. (NBD, but also worth noting that `?vector=false` will of course still result in VECTOR_MODE=true with this code).

src/SpritesheetBuilder.js
```diff
@@ -0,0 +1,134 @@
...
+
+    this.moveQueue.push([moveKey, character, move, moveFrames]);
+    if (this.moveQueue.length === 1) {
+      this.startQueue();
+    }
+  }
+
+  startQueue() {
+    this.p5._incrementPreload();
+    this.onStop = () => this.p5._decrementPreload();
```
Bjvanminnen: Is there a reason we create `onStop` on demand here, rather than just having it be a class method (i.e. sibling of `startQueue`)

src/SpritesheetBuilder.js
```diff
@@ -0,0 +1,134 @@
...
+    // const moveName = window.nativeAPI.world.MOVE_NAMES[move].name;
+    // const svgUrl = `assets/${character.toLowerCase()}.svg#${moveName}`;
+    const svgUrl = `assets/${character.toLowerCase()}.svg`;
+
+    const rasterize = () => {
+      console.count('rasterize');
+      this.context.clearRect(0, 0, CACHED_SIZE * 6, CACHED_SIZE * 4);
+      for (let frame = 0; frame < moveFrames; frame++) {
+        let row = Math.floor(frame / 6);
+        let col = frame % 6;
```
Bjvanminnen: supernit: could be consts

src/SpritesheetBuilder.js
```diff
@@ -0,0 +1,134 @@
...
+    this.context = this.canvas.getContext('2d');
+    setTimeout(this.processQueue, 0);
+  }
+
+  processQueue() {
+    if (this.moveQueue.length <= 0) {
+      console.log('done');
+      delete this.context;
+      delete this.canvas;
+      delete this.cachedSvgImg;
```
Bjvanminnen: should we just do these deletes in `onStop` instead of here?

src/SpritesheetBuilder.js
```diff
@@ -0,0 +1,134 @@
...
+      this.cachedSvgImg.onload = rasterize;
+      this.cachedSvgImg.onerror = skip;
+      this.cachedSvgImg.src = svgUrl;
+    } else {
+      rasterize();
+    }
+  }
+}
+
+module.exports = SpritesheetBuilder;
```
Bjvanminnen: Some general feedback: I'm not sure it makes sense for this to be a class. Seems we might instead be able to just export a `getSpriteSheet`.

If we do want to have this as a class, it might make sense to move local functions like `rasterize` to also be class functions. I found the logic flows here slightly difficult to parse in general, and I think one thing I struggled with was why some functions were defined locally vs. on the class.

I think another thing that would make a lot of this easier to follow would be sticking some param/return js-doc tags on methods.

src/SpritesheetBuilder.js
islemaster: Good call.  This class didn't own `onStop` earlier, now it does we should move these.

islemaster:
Closing for now - pivoting away from this approach per discussion today.

Issue #75, opened 2018-10-30
"Follows" block needs polishing


ryansloan:
Can we ease these? Or make them relative movements instead of absolute?

ryansloan:
These changes from #114 are a lot better, but after trying some different songs I think we want to do a bit more with a similar approach. Hadi has some good ideas about how this should work, we should get his input when we have prototypes. Here's the proposal:
* The change to map the output range to a restricted input range helped a lot, but outliers are still skewing the mean. We should **select a "typical" range for each song and then use that as the domain for the mapping**. We can either do this manually, or compute it like we are doing now (let's talk about options). For most songs*, measures 5-12 work as a "typical" domain.
* The other issue is that the changes should be **relative to the sprite's initial property value** at the time when the mapping is applied. So with this code, both sprites are the same size the whole time:
<img width="418" alt="screen shot 2018-11-07 at 5 53 19 pm" src="https://user-images.githubusercontent.com/1669271/48172820-1cce2f80-e2b6-11e8-8506-3504a760ecda.png">


*The measures 5-12 time span doesn't work well for Bruno Mars, will.i.am, yolanda be cool, Aha.

ryansloan:
I think we need to push this to post-Launch. It will not be a breaking change for any student code.

Bjvanminnen:
Right now, the "following" block doesn't do anything beat related. It's perhaps an interesting idea to have the ability to follow the beat in addition to treble/mid/bass, though I'm not sure exactly what that would look like (do we treat beats as binary data points, or do they have an amplitude for example. how do we smooth between them)

breville:
My prototype wasn't necessarily tied to beats, it was just a way to manually trigger peaks, with the follow-on idea that we could then smooth those peaks (particularly their tails) to create the ongoing "following" behavior.  From what I can tell, the expectation for "following" is that the value peaks fairly quickly at a bass, mid, or treble "hit" and then fades down again.  I think that would look pretty good.

simplycloud:
confirmed fix. resolving.

breville:
Reopening this, even though it's obviously very late, simply because I think the "follows" block really isn't working as it should in almost any situation.

The goal seems simple: the character should "pulse" in time with an easily discernible component of the song.  

After going through a bunch of songs, very rarely does this happen, and even when it kind of does, it's lagging behind the song, in contrast to the tight timing of the dance movements etc.

simplycloud:
[...]

PR #114, opened 2018-11-05
remap energy data

Bjvanminnen:
https://github.com/code-dot-org/dance-party/issues/75

We have a block that allows you to make dancers "follow" particular frequency ranges in the music (bass/mid/treble). In many cases there wasn't enough movement for things to be interesting - i.e. we would have some small amount of variation amongst high values in the bass band.

This PR remaps the energy values using a process I describe below. It does that mapping at runtime when you hit play. Ideally, we would actually do this as part of our preRender step, and it would be easy enough to port this code to the preRender pipeline, but that would also mean needing to regenerate a number of new .json files. After discussing with Ryan and Josh, we decided to stick with this approach (willing to be convinced it is wrong).

The Process:
For each frequency band (bass/mid/treble), calculate the mean value, ignoring those entries where the value is zero, and also the standard deviation.

Come up with a new target range that is within N standard deviations of the mean on either side. Clamp values outside of this range to be on the edge (i.e. if our range is 100-200, then value 50 becomes value 100), then normalize back to using the entire range (such that value 50, which got clamped to 100 is now normalized to 0).

### DIFF
--- a/src/modifySongData.js
+++ b/src/modifySongData.js
@@ -1,0 +1,81 @@
+function clamp(min, max, val) {
+  return Math.max(
+    min,
+    Math.min(max, val)
+  );
+}
+
+function calculateMean(vals) {
+  let sum = vals.reduce((sum, cur) => sum + cur, 0);
+  return sum / vals.length;
+}
+
+/**
+ * This method takes an input list of energy values, and outputs a new list. THe
+ * methodology for figuring out the new list is to calculate the mean and standard
+ * deviation of the input values (ignoring zeros). We then define a new range
+ * consisting of some number of standard deviations in either direction from the
+ * mean. Old values are then clamped to our new range, and normalized. Finally
+ * we do some smoothing between frames, based on our smooth factor
+ * @param {number[]} energy - A list of energy values for a single frequency band
+ *   (i.e. bass, mid, high) with values ranging from 0, 255
+ * @param {number} numDeviations - How many standard deviations in either direction
+ *   from the mean we want to include in our new range
+ * @param {number} smoothFactor - What percentage of previous frames to average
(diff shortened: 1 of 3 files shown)


src/modifySongData.js
```diff
@@ -0,0 +1,84 @@
...
+  // One thing to note with the approach we've chosen: each song/energy band is
+  // independently normalized around energy 128. This means even if a song is
+  // very base heavy, the average bass and average treble will end up being the
+  // same.
+  const bass = getFrequencyEnergy(analysis.map(x => x.energy[0]), numDeviations, smoothFactor);
+  const mid = getFrequencyEnergy(analysis.map(x => x.energy[1]), numDeviations, smoothFactor);
+  const treble = getFrequencyEnergy(analysis.map(x => x.energy[2]), numDeviations, smoothFactor);
+
+  // Create a new analysis that duplicates each frame, but replaces energy values
+  // with out new ones
```
breville: nit: our*

src/modifySongData.js
```diff
@@ -0,0 +1,84 @@
...
+}
+
+/**
+ * Smooths out an array frame by frame. Each new frame consists of smoothFactor
+ * percent of the previous frame and (1-smoothFactor) percent of the current frame
+ * This method mutates the input rg
+ * @param {number[]} rg
+ * @param {number} smoothFactor
+ */
+function smooth(rg, smoothFactor) {
```
Hamms: I'd be interested in seeing what this looks like with and without smoothing, and maybe also seeing a comment explaining why this part of the process is desired.

Bjvanminnen: I'll also add a comment to the code, but this is done so that movement doesnt end up being quite as jerky frame by frame (a big complaint from Marina).

Issue #115, opened 2018-11-05
Integration test fails intermittently


Bjvanminnen:
`node test/integration/spriteDanceTest.js`

I've seen this test fail randomly a couple times in travis. I ran the command 5x locally against master, and it failed once. Would be good to get this passing consistently (or possibly disable the test if that's not an option).

```
>node test/integration/spriteDanceTest.js
TAP version 13
# Dance collision: every measures - rarer move second
ok 1 should be truthy
# Dance collision: every measures - rarer move first
not ok 2 should be truthy
  ---
    operator: ok
    expected: true
    actual:   false
    at: attempt (/Users/brent/git/dance-party/test/integration/spriteDanceTest.js:16:11)
    stack: |-
      Error: should be truthy
          at Test.assert [as _assert] (/Users/brent/git/dance-party/node_modules/tape/lib/test.js:224:54)
          at Test.bound [as _assert] (/Users/brent/git/dance-party/node_modules/tape/lib/test.js:76:32)
          at Test.assert (/Users/brent/git/dance-party/node_modules/tape/lib/test.js:342:10)
          at Test.bound [as true] (/Users/brent/git/dance-party/node_modules/tape/lib/test.js:76:32)
          at attempt (/Users/brent/git/dance-party/test/integration/spriteDanceTest.js:16:11)
          at DanceParty.onPuzzleComplete [as onPuzzleComplete_] (/Users/brent/git/dance-party/test/helpers/runLevel.js:14:7)
          at DanceParty.fail (/Users/brent/git/dance-party/src/p5.dance.js:144:10)
          at Object.eval [as validationCallback] (eval at onInit (/Users/brent/git/dance-party/test/helpers/runLevel.js:24:34), <anonymous>:8:23)
          at DanceParty.draw (/Users/brent/git/dance-party/src/p5.dance.js:946:16)
          at p5Inst.draw (/Users/brent/git/dance-party/src/p5.dance.js:111:32)
  ...
# Dance collision: every seconds - rarer move second
ok 3 should be truthy
# Dance collision: every seconds - rarer move first
```

joshlory:
@epeach are you still tracking this?

https://codedotorg.slack.com/archives/CC2QHV7SR/p1541002217194100

epeach:
Closing for now. First set of test updates are merged in. Will re-open if issue persists.

joshlory:
Still seems to be flakey, even with the latest fix:

[...]

PR #116, opened 2018-11-05
Check ranges for integration tests

epeach:
Addresses: https://github.com/code-dot-org/dance-party/issues/115
Update the validation code so that it is checking a range of values to account for frame rates.

I'll continue to keep an eye on the performance of these tests to understand if this change is effective, but it seems to be working for me locally when run a handful of times.

### DIFF
--- a/levels/spriteDance.js
+++ b/levels/spriteDance.js
@@ -8,11 +8,11 @@ module.exports = {
         lead_dancer = makeNewDanceSprite("CAT", lead_dancer, {x: 200, y: 200});
       });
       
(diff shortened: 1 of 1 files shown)


joshlory:
Did this test consistently fail when cranking the bpm up to >10,000 and pass with these fixes?

epeach:
One of the tests in this suite just failed again for me on Travis. Going to continue investigation.

epeach:
@joshlory I couldn't force a fail on these tests before the changes with bpm >10,000

Issue #91, opened 2018-11-01
Make the "Measure:" string translateable on the canvas


boatmarina:
This string is not currently translateable and should be given how prominent it is:
![image](https://user-images.githubusercontent.com/33850843/47868756-4b35a180-ddc2-11e8-8bd7-3696c12f1d17.png)

joshlory:
Should be resolved with PR https://github.com/code-dot-org/code-dot-org/pull/25849.

PR #117, opened 2018-11-05
Make it possible to translate text in dance party

joshlory:
Include a fallback, for back-compat until we ship the corresponding code-dot-org change.


joshlory:
First half of issue https://github.com/code-dot-org/dance-party/issues/91.

PR #118, opened 2018-11-06
fix songMetadata_ references in draw()

cpirich:
While in `draw()`, we can't assume that `this.songMetadata_ ` has been populated. Fixed this race condition.

### DIFF
--- a/src/p5.dance.js
+++ b/src/p5.dance.js
@@ -945,13 +945,15 @@ module.exports = class DanceParty {
   draw() {
     this.updateEvents_();
 
+    const { bpm, artist, title } = this.songMetadata_ || {};
+
     const context = {
       isPeak: this.peakThisFrame_,
       centroid: this.centroid_,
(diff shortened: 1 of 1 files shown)


joshlory:
Can you link to the PR that regressed this?  Thoughts on reverting that PR, vs. fixing forward?

cpirich:
This broke in this commit: https://github.com/code-dot-org/dance-party/commit/6d45b4566c3f0bcee0e4b99b13c9452b50de135c

src/p5.dance.js
```diff
@@ -945,13 +945,15 @@ module.exports = class DanceParty {
   draw() {
     this.updateEvents_();
 
+    const { bpm, artist, title } = this.songMetadata_ || {};
```
Bjvanminnen: nice approach

cpirich:
https://github.com/code-dot-org/dance-party/pull/82 was the issue

Issue #71, opened 2018-10-30
When using group blocks, "random" option should be the same for all sprites in the group


ryansloan:
Repro: The following code results in all sprites doing different moves
![image](https://user-images.githubusercontent.com/1669271/47738838-3f66a580-dc31-11e8-897e-a6f0c47073ca.png)

Expected: all sprites do the same "random" move.

PR #121, opened 2018-11-06
Same Random Dance for Group

epeach:
https://github.com/code-dot-org/dance-party/issues/71

When a group of dancers is assigned 'random', we change all dancers to the same random group.

Includes tests.


test/unit/spriteManipulation.js
```diff
@@ -393,3 +393,68 @@ test('setPropRandom set sprite y properties between 50 and 350', async t => {
...
+      nativeAPI.setAnimationSpriteSheet("CAT", i, {}, () => {});
+      nativeAPI.world.MOVE_NAMES.push({
+        name: `move${i}`,
+        rest: i == 0,
+      })
+    }
+    nativeAPI.world.fullLengthMoveCount = moveCount;
+    nativeAPI.world.restMoveCount = 1;
+
+    const sprite = nativeAPI.makeNewDanceSprite("CAT", null, {x: 200, y: 200});
```
joshlory: Tiny style nit: why `{ moveCount = 1, testCode }` elsewhere, but no spaces inside the braces on `{x: 200, y: 200}`?

Would be good to get eslint running on the test folder as well...

src/p5.dance.js
```diff
@@ -356,42 +356,53 @@ module.exports = class DanceParty {
 
 // Dance Moves
 
-  changeMoveLR(sprite, move, dir) {
-    if (!this.spriteExists_(sprite)) {
-      return;
-    }
+  checkDanceMoveCount() {
```
joshlory: Not used?

test/unit/spriteManipulation.js
epeach: Split into new PR: https://github.com/code-dot-org/dance-party/pull/134

src/p5.dance.js
```diff
@@ -356,42 +356,49 @@ module.exports = class DanceParty {
 
 // Dance Moves
 
-  changeMoveLR(sprite, move, dir) {
-    if (!this.spriteExists_(sprite)) {
-      return;
-    }
+  getNewChangeMove(requested_change, current_move) {
```
Bjvanminnen: lets stick some @param tags on here

also isnt the style usually requestedChange and currentMove in our JS code?

src/p5.dance.js
```diff
@@ -356,42 +356,49 @@ module.exports = class DanceParty {
...
     // Number of valid full length moves
     const { fullLengthMoveCount, restMoveCount } = this.world;
     const firstNonRestingMoveIndex = restMoveCount;
     // The "rest" moves are assumed to always be at the beginning
     const nonRestingFullLengthMoveCount = fullLengthMoveCount - restMoveCount;
+    if (nonRestingFullLengthMoveCount <= 1) {
+      throw "next/prev/rand requires that we have 2 or more non-resting full length moves";
```
Bjvanminnen: in general, it's better practice to throw errors rather than strings, i.e. `throw new Error('next/prev/rand...');`.

I think this is in part because a lot of error handlers expect to get an error object instead. I think it might also mean better call stack reporting, but I might be wrong about that.

src/p5.dance.js
```diff
@@ -356,42 +356,49 @@ module.exports = class DanceParty {
...
+    } else if (requested_change === "prev") {
+      move = current_move - 1;
+      if (move < firstNonRestingMoveIndex) {
+        move = fullLengthMoveCount - 1;
+      }
+    } else if (requested_change === "rand") {
+      // Make sure random switches to a new move
+      while (move === current_move) {
+        move = randomInt(this.world.restMoveCount, this.world.fullLengthMoveCount - 1);
+      }
```
Bjvanminnen: One way to do this and avoid a while loop is to come up with a random number between 1 and the totalLength - 1, and then increment by that number.

i.e. if you want a random number in the range (0, 10) and your current index is i, you would do
```
length = 10; // 11 - 1
iNext = (i + randomInt(1, length)) % length
```

The math to get this approach working for your scenario would be a little more difficult, since the bottom of your range is non-zero, but it could be an option if you wanted to avoid the while loop (which I always find a little scary).

test/unit/groupTest.js
```diff
@@ -93,6 +93,46 @@ test('changing dance moves for all cats updates only all cat dancers', async t =
...
+      name: `move${i}`
+    })
+  }
+
+  nativeAPI.world.fullLengthMoveCount = moveCount;
+  nativeAPI.world.restMoveCount = 1;
+
+  let spriteGroup = [];
+  const groupCount = 5;
+  for(let i = 0; i < groupCount; i++){
```
Bjvanminnen: going to get some lint errors on your for loops i think once this is merged with your eslint changes

test/unit/groupTest.js
```diff
@@ -93,6 +93,46 @@ test('changing dance moves for all cats updates only all cat dancers', async t =
...
+
+  //Initial
+  t.equal(spriteGroup[0].current_move, 0);
+
+  nativeAPI.changeMoveEachLR('all', 'rand');
+
+  const newMove = spriteGroup[0].current_move;
+  for(let i = 0; i < groupCount; i++){
+    t.notEqual(spriteGroup[i].current_move, 0);
+    t.equal(spriteGroup[i].current_move, newMove);
```
Bjvanminnen: Because the current_move going in was 0, and our restMoveCount was 1, our rand function was never going to give us a 0 regardless.

I wonder if we should also test with a starting current_move > restMoveCount

test/unit/spriteManipulation.js
```diff
@@ -393,3 +393,68 @@ test('setPropRandom set sprite y properties between 50 and 350', async t => {
...
+  // Verify invalid number of tests:
+  await subTest({ testCode: ({nativeAPI, sprite}) => {
+    let error = null;
+    try {
+      // Requesting 'rand' when only one dance should fail
+      nativeAPI.getNewChangeMove(sprite, 'rand', 1);
+    } catch (e) {
+      error = e;
+    }
+    t.notEqual(error, null, "next/prev/rand requires that we have 2 or more non-resting full length moves");
```
Bjvanminnen: I think we could use `t.throws`
https://github.com/substack/tape#tthrowsfn-expected-msg

epeach: Spent some time investigating this, but abandoned after a while to not get blocked by it.

epeach:
@Bjvanminnen PTAL, thanks!

src/p5.dance.js
```diff
@@ -361,42 +361,54 @@ module.exports = class DanceParty {
...
+   * @param {number} currentMove - value representing current move of sprite
+   */
+  getNewChangedMove(requestedChange, currentMove) {
     // Number of valid full length moves
     const { fullLengthMoveCount, restMoveCount } = this.world;
     const firstNonRestingMoveIndex = restMoveCount;
     // The "rest" moves are assumed to always be at the beginning
     const nonRestingFullLengthMoveCount = fullLengthMoveCount - restMoveCount;
+    if (nonRestingFullLengthMoveCount <= 1) {
+      throw new Error ("next/prev/rand requires that we have 2 or more non-resting full length moves");
```
Bjvanminnen: supernit: here and elsewhere, I wouldnt expect a space after Error

Issue #98, opened 2018-11-01
stops following block doesnt seem to work


Bjvanminnen:
![image](https://user-images.githubusercontent.com/1767466/47885142-d3cb3680-ddf0-11e8-8828-8c7a7f92656c.png)

I would expect rotation to be fixed here, but it is not

PR #136, opened 2018-11-06
Rework behaviors

Bjvanminnen:
https://github.com/code-dot-org/dance-party/issues/98

This is kind of a big change for a smallish bug. It might be that the right short-term decision is make a smaller (but hackier) fix, but I want to at least capture my thoughts/knowledge of the state of the world here.

## Behaviors
We have this concept of behaviors. They are essentially callbacks that are attached to a specific sprite and are called once per draw call starting after frame number 2.

Behaviors can be attached/removed to sprites.

Behaviors exist in two forms. The first is a function that (I believe) updates the animation frame of the sprite. This behavior gets attached to every single sprite and is never removed.

The second form is behaviors that look at the current energy levels in the song and update a sprite's properties. These behaviors get added or removed by users using specific blocks (though removal does not currently work).

## The bug
When seeing a "start mapping" block, we would create a new Behavior along with a new function that would be called per draw call. This function would update the sprite properties.

When seeing a "stop mapping" block, we would look for the equivalent Behavior. The problem is that one of the things we would look at when seeing if two Behaviors are equal is whether they have the same function. They never would, see we created a new function for every new Behavior instance.

## The fix
Simplify the concept of behaviors to essentially only be those things that the user adds and removes. For the first class of behavior, instead just call that code directly (instead of wrapping it in a behavior). This also means we can make our Behavior class (now called `EnergyBehavior` less generic, and less likely to have bugs.

## The catch
Some of our existing levels check `sprite.behaviors.length` in their validation. We could fix all these up, but making changes like that this late feels a little risky. Instead I've created a NoopBehavior so that count remains the same. Incidentally, really we should probably have our validation code doing something like `nativeAPI.getNumberOfBehaviors()` instead of directly accessing sprite.behaviors.

TODO: Add tests before merging.

### DIFF
--- a/src/p5.dance.js
+++ b/src/p5.dance.js
@@ -4,12 +4,36 @@ const P5 = require('./loadP5');
 const Effects = require('./Effects');
 const replayLog = require('./replay');
 
-function Behavior(func, extraArgs) {
-  if (!extraArgs) {
-    extraArgs = [];
+class Behavior {
+  constructor(func, extraArgs=[]) {
+    this.func = func;
+    this.extraArgs = extraArgs;
   }
-  this.func = func;
-  this.extraArgs = extraArgs;
+
+  execute(sprite) {
+    this.func.apply(null, [sprite].concat(this.extraArgs));
+  }
+
+  equals(other) {
+    if (this.func.name && other.func.name) {
+      // These are legacy behaviors, check for equality based only on the name.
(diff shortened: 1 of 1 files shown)


Bjvanminnen:
Closing in favor of https://github.com/code-dot-org/dance-party/pull/140. Might be worth a revisit post-HOC

PR #140, opened 2018-11-06
Properly remove behaviors when we stopMapping

Bjvanminnen:
Fix for https://github.com/code-dot-org/dance-party/issues/98

See https://github.com/code-dot-org/dance-party/pull/136 for analysis and a bigger fix we could take instead.

I think 136 + removal of the NoopBehavior hack mentioned there is the right long term fix for this bug. However, that's also a lot of change this short before HOC. This PR is a smaller scoped fix that I think is probably the right one to make short term.

Here we depend on coming up with an identifier per behavior that consists of the property we're wanting to modify and the energy range we want to track. Using this identifier we can then easily add/remove behaviors.

This is a reasonably good fix, but unlike #136 doesn't address the fact that we're using the Behavior concept for two reasonably different things.

### DIFF
--- a/src/p5.dance.js
+++ b/src/p5.dance.js
@@ -794,6 +794,9 @@ module.exports = class DanceParty {
     }
 
     const index = sprite.behaviors.findIndex(b => b.id === behaviorId);
+    if (index === -1) {
+      return;
+    }
     sprite.behaviors.splice(index, 1);
   }
 
@@ -829,6 +832,7 @@ module.exports = class DanceParty {
 
   stopMapping(sprite, property, range) {
     const id = [property, range].join('-');
+    console.log('id: ', id);
     this.removeBehavior_(sprite, id);
   }
 
--- a/src/p5.dance.js
+++ b/src/p5.dance.js
@@ -832,7 +832,6 @@ module.exports = class DanceParty {
 
   stopMapping(sprite, property, range) {
     const id = [property, range].join('-');
-    console.log('id: ', id);
     this.removeBehavior_(sprite, id);
   }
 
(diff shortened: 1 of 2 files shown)


Bjvanminnen:
cc @mrjoshida Who I suspect initially wrote the behavior code and might see gotchas that I missed.

src/p5.dance.js
```diff
@@ -845,32 +826,14 @@ module.exports = class DanceParty {
...
-      } else if (property === "tint") {
-        energy = Math.round(this.p5_.map(energy, 0, 255, 0, 360));
-        energy = "hsb(" + energy + ",100%,100%)";
-      }
-      sprite[property] = energy;
-    }, [property, range]);
-    //behavior.func.name = "mapping" + property + range;
-    this.removeBehavior_(sprite, behavior);
+    const id = [property, range].join('-');
+    console.log('id: ', id);
```
islemaster:
```suggestion
```

test/unit/spriteManipulation.js
```diff
@@ -393,3 +393,46 @@ test('setPropRandom set sprite y properties between 50 and 350', async t => {
...
+  t.equal(sprite.behaviors.length, 1);
+
+  nativeAPI.startMapping(sprite, 'x', 'bass');
+
+  t.equal(sprite.behaviors.length, 2);
+
+  // adding the same mapping again gets ignored
+  // TODO: this fails
+  // nativeAPI.startMapping(sprite, 'x', 'bass');
+  // t.equal(sprite.behaviors.length, 2);
```
islemaster: I don't know much about behaviors, is this something we need to address before merge?

Bjvanminnen: bah, that should be uncommented (and will hopefully pass)

Bjvanminnen:
I also prefer 136, but am more scared of breaking something with that change.

Issue #83, opened 2018-10-31
set width/height doesn't seem to work


Bjvanminnen:
![image](https://user-images.githubusercontent.com/1767466/47805647-10b20300-dcf5-11e8-9727-da3cf32cce4f.png)

Setting size works fine, but setting width/height doesn't seem to accomplish anything.

ryansloan:
Not sure when this regressed...

PR #163, opened 2018-11-07
Width height Prop fix

epeach:
Addresses: https://github.com/code-dot-org/dance-party/issues/83

Similar to when we initialize p5 in Game Lab, turning on this sprite permits p5 to scale the sprite horizontally or vertically when the height and width are set. https://github.com/code-dot-org/p5.play/blob/4ae6d3dc336b2325fdb6d8816ee3f1ce95370d23/lib/p5.play.js#L1783

Includes test to make sure we don't remove the flag

Paired with @islemaster

### DIFF
--- a/src/p5.dance.js
+++ b/src/p5.dance.js
@@ -104,6 +104,7 @@ module.exports = class DanceParty {
       p5Inst.preload = () => this.preload();
       p5Inst.setup = () => this.setup();
       p5Inst.draw = () => this.draw();
+      //Allows the sprite width and height to be set independently
       this.p5_._fixedSpriteAnimationFrameSizes = true;
     }, container);
   }
--- a/src/p5.dance.js
+++ b/src/p5.dance.js
@@ -104,6 +104,7 @@ module.exports = class DanceParty {
       p5Inst.preload = () => this.preload();
       p5Inst.setup = () => this.setup();
       p5Inst.draw = () => this.draw();
+      this.p5_._fixedSpriteAnimationFrameSizes = true;
     }, container);
   }
 
--- a/test/unit/spriteManipulation.js
+++ b/test/unit/spriteManipulation.js
@@ -385,3 +385,12 @@ test('setPropRandom set sprite y properties between 50 and 350', async t => {
 
   nativeAPI.reset();
 });
+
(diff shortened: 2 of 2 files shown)


src/p5.dance.js
```diff
@@ -104,6 +104,8 @@ module.exports = class DanceParty {
       p5Inst.preload = () => this.preload();
       p5Inst.setup = () => this.setup();
       p5Inst.draw = () => this.draw();
+      //Allows the sprite width and height to be set independently
+      this.p5_._fixedSpriteAnimationFrameSizes = true;
```
Bjvanminnen: Is this documented somewhere? Having a name starting with `_` usually means it should be treated as private, and we shouldnt really be going in and mucking with it. It would be better if p5 had some sort of API we could go through to target this.

test/unit/spriteManipulation.js
```diff
@@ -385,3 +385,12 @@ test('setPropRandom set sprite y properties between 50 and 350', async t => {
...
 });
+
+test('p5 fixedSpriteAnimationSizes is true', async t => {
+  //Turning this flag on allows us to scale a sprite horizontally and vertically
+  const nativeAPI = await helpers.createDanceAPI();
+  t.ok(nativeAPI.p5_._fixedSpriteAnimationFrameSizes);
+  t.end();
+
+  nativeAPI.reset();
+});
```
Bjvanminnen: can we also have a test that makes sure that changing width/height now works?

epeach: Before this change, the sprite would accurately save width and height as expected, but p5 wouldn't use the width and height while drawing to ensure the width/height ratio remained the same. I'll see what I can do to add a test, but it will be non-trivial to have a regression test.

src/p5.dance.js
epeach: No documentation on this, but it's a practice I borrowed from GameLabP5: https://github.com/code-dot-org/code-dot-org/blob/30a0a9814f123b72a16661ecc9d852640dc31f1c/apps/src/gamelab/GameLabP5.js#L193

Might be worth adding a function to the p5 API, but I think that can come in a follow-up PR.

epeach:
@Bjvanminnen PTAL - thanks!

Issue #165, opened 2018-11-07
"Every N measures" blocks interfere with each other


breville:
Repro:

The "every 2 measures" block never fires at the same time as the "every 4 measures" block.  Surely it should.

![screenshot 2018-11-07 19 50 45](https://user-images.githubusercontent.com/2205926/48120579-a1c14500-e2c6-11e8-9552-17b5ff447f4b.png)

The duck should move every time the moose moves.  It doesn't because its event gets interfered with.

breville:
It appears that this might be by design:

https://github.com/code-dot-org/dance-party/blob/513c981724f5e6f3fab563eb5c507fbc85148423/src/p5.dance.interpreted.js#L67

In which case the priority is how many measures/seconds are specified in the event:

https://github.com/code-dot-org/dance-party/blob/513c981724f5e6f3fab563eb5c507fbc85148423/src/p5.dance.interpreted.js#L136

But that doesn't make sense for a student, does it?

epeach:
The fix I had made for this earlier issue appears to have oversimplified the situation and not accounted for this situation.

joshlory:
Weirdly, I'm not able to repro this in an integration test: https://github.com/code-dot-org/dance-party/compare/every-measure-collision-test

ryansloan:
I realized this was an ambiguous spec on my part. I'll try again, let me know if there are any questions or confusion! 
Here's expected behavior:
* Run every applicable event
* Run them in the following order: 
1) `Every N seconds`
2) `Every N measures`
3) `At N seconds`
4) `At N measures`

Within a group (ie: for the `every 2 measures` and `every 4 measures` example, we should execute them in order based on Y position (as is the blockly standard), starting from the top and moving down.

The last item to be executed can potentially override earlier items. For example, with this code:
<img width="357" alt="screen shot 2018-11-07 at 12 32 38 pm" src="https://user-images.githubusercontent.com/1669271/48159130-415fe280-e289-11e8-8d09-7d4a0d9883d3.png">

The `after 8 measures` block completely tramples over the `every 4 measures` block because it is executed last.

joshlory:
@ryansloan I'm confused by this comment:

> we should execute them in order based on Y position (as is the blockly standard), starting from the top and moving down

Wasn't issue https://github.com/code-dot-org/dance-party/issues/43 about code position affecting event ordering?

ryansloan:
@joshlory good point, sorry, there's still some ambiguity, fixing.

[...]

PR #166, opened 2018-11-07
After/Every N measures/seconds: generate events at proper times

### LINKED ISSUE #51: "after N measures" and "every N measures" events start too early
Current: they begin in measure N
Expected: they begin in measure N+1

breville:
Previously, `"Every 4 measures"` was generating events at the beginning of measures `1, 4, 8, 12`.   Now, they're at the beginning of measures `5, 9, 13`.

`"Every 4 seconds"` was generating events at the beginning of seconds `0, 4, 8, 12`.  Now they're at the beginning of seconds `4, 8, 12`.

Fractional values should also work.  `"Every 0.25 measures"` will be `1.25, 1.5, 1.75`.  `"Every 0.25 seconds"` will be `0.25, 1.5, 1.75`.

Finally, the "After 4 measures" event (which uses the deceptively-named `atTimestamp` function) was generating an event at the beginning of measure 4.  Now it's at the beginning of measure 5.

Fixes #51

### DIFF
--- a/levels/spriteDance.js
+++ b/levels/spriteDance.js
@@ -7,13 +7,14 @@ module.exports = {
       whenSetup(function () {
         lead_dancer = makeNewDanceSprite("CAT", lead_dancer, {x: 200, y: 200});
       });
-      
+
       everySeconds(2, "measures", function () {
         changeMoveLR(lead_dancer, 3, -1);
       });
-      
+
       everySeconds(4, "measures", function () {
         changeMoveLR(lead_dancer, 1, -1);
+        setProp(lead_dancer, "scale", 50);
       });
     `,
       `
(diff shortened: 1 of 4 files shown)


levels/spriteDance.js
```diff
@@ -7,13 +7,14 @@ module.exports = {
...
-      
+
       everySeconds(2, "measures", function () {
         changeMoveLR(lead_dancer, 3, -1);
       });
-      
+
       everySeconds(4, "measures", function () {
         changeMoveLR(lead_dancer, 1, -1);
+        setProp(lead_dancer, "scale", 50);
```
breville: Spent a bunch of time trying to fire this event in the event immediately above, before discovering https://github.com/code-dot-org/dance-party/issues/165.

src/p5.dance.interpreted.js
```diff
@@ -121,13 +125,27 @@ function atTimestamp(timestamp, unit, event) {
 }
 
 function everySeconds(n, unit, event) {
```
Bjvanminnen: Would  be great to start getting some param/return tags on these functions so that they can be more easily understood going forward

PR #172, opened 2018-11-07
pass result of playSound_ callback to play callback

davidsbailey:


davidsbailey:
This is just one piece of plumbing needed to improve error handling when playing songs. The rest of the work will be in the code.org repo. This is a backward compatible change.

PR #184, opened 2018-11-07
inject interpreted code into demo

Bjvanminnen:


demo.js
```diff
@@ -1,12 +1,61 @@
...
+  }
+  params.push(code);
+  const ctor = function () {
+    return Function.apply(this, params);
+  };
+  ctor.prototype = Function.prototype;
+  const result = new ctor().apply(null, args);
+
+  // inject onto window
+  functionNames.forEach(name => window[name] = result[name]);
```
Bjvanminnen: potentially contentious?

could also stick it onto something like `window.interpreted.atTimestamp`

joshlory: Yeah, I think I like the parallel between `nativeAPI.` and `interpretedAPI.`

demo.js
```diff
@@ -1,12 +1,61 @@
...
+  for (let k of Object.keys(globals)) {
+    params.push(k);
+    args.push(globals[k]);
+  }
+  params.push(code);
+  const ctor = function () {
+    return Function.apply(this, params);
+  };
+  ctor.prototype = Function.prototype;
+  const result = new ctor().apply(null, args);
```
joshlory: Worth de-duping this against `runLevel.js`?  I like the `return {${functionNames.join(',')}};` pattern, we could do the same for integration tests.

https://github.com/code-dot-org/dance-party/blob/dffdf04b6a1a7f28c0509660c2dde7faf85e888d/test/helpers/runLevel.js#L27-L38

joshlory: Ah I see, the `require('fs')` is needed in NodeJS, and `require('raw-loader!...)` is needed in the browser 😕.

demo.js
Bjvanminnen: Looking at this further, I think it might not be worthwhile to expose these functions anywhere. It looks like if I add something like `atTimestamp` I still need to call `getCueList` afterwards to get things working properly.

It might make more sense to just say that if you want to step through custom code, you stick it in the `// Sample user code:` section of demo.js rather than trying to do things from the chroms console.

PR #187, opened 2018-11-07
Update demo playground: run any Dance project by pasting the "Show Code" output

joshlory:
One idea to make the playground a little more useful.  Leans heavily on PR https://github.com/code-dot-org/dance-party/pull/184 to make this possible!

screen shot 2018-11-07 at 2 52 52 pm

### DIFF
--- a/demo.js
+++ b/demo.js
@@ -3,6 +3,8 @@ import jazzy_beats from './metadata/jazzy_beats';
 import interpreted from 'raw-loader!./src/p5.dance.interpreted.js';
 import injectInterpreted from './test/helpers/injectInterpreted';
 
+const textareaCode = document.querySelector('#code');
+
 const nativeAPI = window.nativeAPI = new DanceParty({
   onPuzzleComplete: () => {},
   playSound: (url, callback, onEnded) => setTimeout(() => {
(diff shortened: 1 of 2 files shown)


Bjvanminnen:
This is super cool!

I didn't realize these methods were also global (i.e. that we can do `setBackground` and not just `nativeAPI.setBackground`.

Should we auto-run the initial contents, so that as I'm developing locally I don't have to hit run every time I make a change?

demo.js
```diff
@@ -9,29 +9,35 @@ const nativeAPI = window.nativeAPI = new DanceParty({
...
   onInit: () => {
+    document.querySelector('#run').style.display = 'inline';
+  },
+  container: 'dance',
+});
+
+document.querySelector('#code').innerText = `makeNewDanceSprite("CAT", null, {x: 200, y: 200});
+
+atTimestamp(2, "measures", function () {
+  nativeAPI.setBackgroundEffect("disco");
```
Bjvanminnen: I guess the nativeAPI is unnecessary here

joshlory:
Added auto-run in commit a9116316364dd350506d6911602278297f2a1902 but it looks like only Firefox keeps the textarea contents around across reloads.  We could do follow-on work to track this in session storage, if desired!

Issue #167, opened 2018-11-07
Dance move mapping incorrect


ryansloan:
"Clap high" mapping to the new short-burst "kick", "roll" maps to clap high. @cpirich I'm not sure if this is a Levelbuilder change I need to make, or if there's a bug in the dance ordering

ryansloan:
We should pair this with https://github.com/code-dot-org/dance-party/issues/178

cpirich:
Fixed and merged in both repos

ryansloan:
@cpirich This doesn't appear to be working in staging

PR #208, opened 2018-11-08
fix sorting of move names

cpirich:
* Sorting of `MOVE_NAMES` was broken (references to `undefined` properties behave differently than `false` properties)
* Added test case to verify sorting. This one was a little tricky as it is our first test case that runs all of the init code (loading JSON, spritesheets, etc.). I tried to stub out `XMLHttpRequest` with sinon, but that proved to be quite complex in our current harness environment. For now, I've done a simpler change which is to stub out `p5`'s `loadJSON()` and `loadImageElement()` - which allows us to run through the `DanceParty` constructor properly with `MOVE_NAMES` and `SPRITE_NAMES` preset.

### DIFF
--- a/src/p5.dance.js
+++ b/src/p5.dance.js
@@ -92,7 +92,7 @@ module.exports = class DanceParty {
     // Sort after spriteConfig function has executed to ensure that
     // rest moves are at the beginning and shortBurst moves are all at the end
     this.world.MOVE_NAMES = this.world.MOVE_NAMES.sort((move1, move2) => (
-      !!move1.rest * -2 * !!move2.rest * 2 + !!move2.shortBurst * -1 + !!move1.shortBurst * 1
+      !!move1.rest * -2 + !!move2.rest * 2 + !!move2.shortBurst * -1 + !!move1.shortBurst * 1
     ));
     this.world.restMoveCount = this.world.MOVE_NAMES.filter(move => move.rest).length;
     this.world.fullLengthMoveCount = this.world.MOVE_NAMES.filter(move => !move.shortBurst).length;
--- a/src/p5.dance.js
+++ b/src/p5.dance.js
@@ -92,7 +92,7 @@ module.exports = class DanceParty {
     // Sort after spriteConfig function has executed to ensure that
     // rest moves are at the beginning and shortBurst moves are all at the end
     this.world.MOVE_NAMES = this.world.MOVE_NAMES.sort((move1, move2) => (
-      move1.rest * -2 * move2.rest * 2 + move2.shortBurst * -1 + move1.shortBurst * 1
+      !!move1.rest * -2 * !!move2.rest * 2 + !!move2.shortBurst * -1 + !!move1.shortBurst * 1
     ));
     this.world.restMoveCount = this.world.MOVE_NAMES.filter(move => move.rest).length;
     this.world.fullLengthMoveCount = this.world.MOVE_NAMES.filter(move => !move.shortBurst).length;
--- a/test/unit/spriteManipulation.js
+++ b/test/unit/spriteManipulation.js
@@ -252,7 +252,36 @@ test('Sprite dance changes will throw with invalid parameters', async t => {
 });
 
 test('Sprite move sorting works reliably', async t => {
-  const moveNames = [
+
+  const P5 = require('../../src/loadP5');
+
+  P5.prototype.loadJSON = function (_url, callback) {
+    setTimeout(() => {
+      this._preloadCount--;
+      this._runIfPreloadsAreDone();
+      callback('{"frames":{}}');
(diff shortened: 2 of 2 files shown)


ryansloan:
Fixes #167

src/p5.dance.js
```diff
@@ -92,7 +92,7 @@ module.exports = class DanceParty {
     // Sort after spriteConfig function has executed to ensure that
     // rest moves are at the beginning and shortBurst moves are all at the end
     this.world.MOVE_NAMES = this.world.MOVE_NAMES.sort((move1, move2) => (
-      move1.rest * -2 * move2.rest * 2 + move2.shortBurst * -1 + move1.shortBurst * 1
+      !!move1.rest * -2 * !!move2.rest * 2 + !!move2.shortBurst * -1 + !!move1.shortBurst * 1
```
islemaster: Shouldn't this be
```suggestion
      !!move1.rest * -2 + !!move2.rest * 2 + !!move2.shortBurst * -1 + !!move1.shortBurst * 1
```
As-is you've got:
image
But it should be:
image

test/unit/spriteManipulation.js
```diff
@@ -251,6 +251,55 @@ test('Sprite dance changes will throw with invalid parameters', async t => {
   t.end();
 });
 
+test('Sprite move sorting works reliably', async t => {
+  const moveNames = [
+    {
+      name: 'rest',
+      rest: true,
+    },
```
islemaster: Please update your test case to have a rest move further down the list, so we're actually checking that rest move sorting works.

test/unit/spriteManipulation.js
```diff
@@ -251,6 +251,55 @@ test('Sprite dance changes will throw with invalid parameters', async t => {
...
+    {
+      name: 'shortBurst2',
+      shortBurst: true,
+    },
+  ];
+
+  const P5 = require('../../src/loadP5');
+
+  P5.prototype.loadJSON = function (_url, callback) {
```
islemaster: Why not `sinon.stub(P5.prototype, 'loadJSON').callsFake(...)` here? Or if this is just to get initialization working, should we do this inside the `createDanceAPI` helper?

test/unit/spriteManipulation.js
```diff
@@ -251,6 +251,55 @@ test('Sprite dance changes will throw with invalid parameters', async t => {
...
+      callback(new Image());
+    }, 0);
+  };
+
+  const nativeAPI = await helpers.createDanceAPI({ moveNames, spriteConfig: world => { world.SPRITE_NAMES = ['foo']; } });
+
+  nativeAPI.p5_.noLoop();
+
+  // The MOVE_NAMES as provided have not been changed.
+  t.deepEqual(nativeAPI.world.MOVE_NAMES, moveNames);
```
islemaster: I worry that this isn't a sufficient test for sorting because we could remove the sorting step entirely and this check would still pass.

src/p5.dance.js
cpirich: thanks for catching that. hence, the need to have more test cases.

test/unit/spriteManipulation.js
cpirich: yes, we really need to have multiple cases. I'm fixing that.
The 1st priority was to make sure that we didn't change a properly sorted list, since that is the only case that we need today within the cdo repo.
Also, because of how levelbuilder is precomputing sort indexes, we really can't afford to resort these when this component is hosted within the cdo repo.
But that seems like a limitation upstream, and I'd like this component to behave in a rational way since the code made assumptions about where the different types of dance moves existed within the array (rests at the beginning, short bursts at the end)

test/unit/spriteManipulation.js
cpirich: `sinon`: while I added while experimenting with it, we don't currently have it in this project. We could add it for this case if you think it helps.

Inside `createDanceAPI`: I'm not sure we're going to want to override these is any consistent way throughout all of the test cases. These specific overrides are just good enough to test the sorting, but nothing else. Maybe we can come up with some great overrides that return fixtures and use them everywhere. Or maybe we'll switch to karma/phantom and use sinon fakeserver.

test/unit/spriteManipulation.js
cpirich: For sure, this is only the one test case I was focused on to regress the bug I'm trying to fix. Don't mess up a pre-sorted array.

test/unit/spriteManipulation.js
islemaster: I added sinon two days ago in https://github.com/code-dot-org/dance-party/pull/161 for testing the measure counter. (Fake timers are really useful for some of the behaviors in this repo).

PR #222, opened 2018-11-08
Changes to event system

Bjvanminnen:
Resolves https://github.com/code-dot-org/dance-party/issues/165.

The first three commits here are just code cleanliness, and could probably be separated entirely if desired.

The fourth commit is the meat of the change. Instead of having a single queue, we have three different event queues. One of these is for "every" events, one is for "after" events, and the last is for all other events. The three events get run in that order. Within each queue, items are prioritized appropriately.

I've some manual testing of this code, and have written unit tests.

### DIFF
--- a/src/p5.dance.interpreted.js
+++ b/src/p5.dance.interpreted.js
@@ -18,6 +18,12 @@ var MOVES = {
   Thriller: 11
 };
 
+var QueueType = {
+  every: 'every',
+  after: 'after',
+  other: 'other'
+};
+
 // Event handlers, loops, and callbacks.
 var inputEvents = [];
 var setupCallbacks = [];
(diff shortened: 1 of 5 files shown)


src/p5.dance.interpreted.js
```diff
@@ -56,25 +65,57 @@ function runUserSetup() {
...
+    // There are ways we could ensure that
+    return a.priority - b.priority;
   }
+
+  function executeFuncs(item) {
+    item.func();
+  }
+
+  queues[QueueType.every].sort(prioritySort).forEach(executeFuncs);
+  queues[QueueType.after].sort(prioritySort).forEach(executeFuncs);
```
Bjvanminnen: I think we could actually get rid of the priority/sort for after

joshlory:
Ok, I think this is ready for review!

src/p5.dance.js
```diff
@@ -291,7 +288,7 @@ module.exports = class DanceParty {
     sprite.looping_move = 0;
     sprite.looping_frame = 0;
     sprite.current_move = 0;
-    sprite.previous_move = 0;
+    sprite.previous_move = 0; // I don't think this is used?
```
epeach: Probably safe to remove, but non-blocking.

src/p5.dance.js
```diff
@@ -895,19 +892,23 @@ module.exports = class DanceParty {
     return this.p5_.allSprites.indexOf(sprite) > -1;
   }
 
+  /**
+   * @return {Object} TODO: describe
```
epeach: Non-blocking, but worth coming back to. The documentation is improving the readability so much - thanks!

test/unit/eventsTest.js
```diff
@@ -0,0 +1,144 @@
...
+    atTimestamp(2, "seconds", function () {});
+  `);
+
+  t.deepEqual(interpretedAPI.getCueList(), {seconds: [2], measures: []});
+
+  t.end();
+  nativeAPI.reset();
+});
+
+test('conflicting everySeconds and atTimestamp cues', async t => {
```
epeach: These tests are great - I wonder if the coverage is sufficient here that we just remove the flakey spriteDanceTest.js?

joshlory: I'm tempted to keep both for now, because these new tests don't cover all the scenarios from the original tests.  Maybe we could do future cleanup to turn the integration tests (that use `runLevel.js`) into unit tests?

src/p5.dance.interpreted.js
```diff
@@ -56,25 +65,57 @@ function runUserSetup() {
   });
 }
 
+/**
+ * @param {Object} events - An object where each key is an event type. Each value
+ *  is another object, where the keys represeent the param of the event to be run
```
breville: nit: represent

src/p5.dance.interpreted.js
```diff
@@ -56,25 +65,57 @@ function runUserSetup() {
...
+ */
 function runUserEvents(events) {
-  var currentEvents = {};
+  // We have three separate event queues.
+  // First we run every N seconds/measures events
+  // Then we run after N seconds/measures events
```
breville: Sounds like this is really going to be "at", not "after", and even if we can't change the user-facing block, it's probably worth having the right nomenclature in the code.

src/p5.dance.interpreted.js
```diff
@@ -56,25 +65,57 @@ function runUserSetup() {
...
+      });
     }
   }
-  for (var input in currentEvents){
-    if(currentEvents.hasOwnProperty(input)){
-      currentEvents[input].event();
-    }
+
+  function prioritySort(a, b) {
+    // TODO: If compareFunction(a, b) returns 0, leave a and b unchanged with respect
```
breville: Is *this* the compareFunction?  Slightly confused by what this comment refers to, and what is left TODO.

src/p5.dance.interpreted.js
```diff
@@ -93,38 +134,60 @@ function ifDanceIs(sprite, dance, ifStatement, elseStatement) {
...
 }
 
-function atTimestamp(timestamp, unit, event) {
+/**
+ * @param {number} timestamp
+ * @param {string} unit - Should be "measures" or "seconds"
+ * @param {function} func - Code to run when event fires
+ */
+function atTimestamp(timestamp, unit, func) {
+  // Despite the functions name, if we call atTimestamp(4, "measures", foo), we
```
breville: Well, we shall see about this separately.

joshlory: Not blocking this PR, I hope!

src/p5.dance.interpreted.js
joshlory: Correct, `prioritySort` is passed to the `.sort` below:

https://github.com/code-dot-org/dance-party/blob/658c9a6334e1c47930efb050b5c95f65421f4612/src/p5.dance.interpreted.js#L116-L117

Chatted with Ryan, he's not worried about the "non-stable sort" case, because we only hit that if a student tries using two different hat blocks to run an event at the exact same time.  Any events with differing priority (or "after" vs. "every") will always be sorted correctly.

src/p5.dance.interpreted.js
joshlory: I'm tempted to leave this as-is because it's correct for the _current_ state of the world.  If we make changes to after/at, we should update this comment at that time.

PR #223, opened 2018-11-08
Add frameRate to test interface for stress testing

islemaster:
I'd like to monitor and report framerate from our stress tests but can't get to p5 right now.  Exposing this in the test interface (which in turn gets exposed on window) so we can check it during stress tests.


islemaster:
Closing in favor of more robust timing metrics in https://github.com/code-dot-org/dance-party/pull/227

Issue #107, opened 2018-11-03
Hypno and Swirl backgrounds break on IE11


boatmarina:
I was able to repro this both in levelbuilder and in prod, both levels 12 and 13.  if you select the hypno background, nothing renders on the canvas:
![image](https://user-images.githubusercontent.com/33850843/47958320-4c173080-df86-11e8-864a-c9a0a7eab6d5.png)

and after that, any time you select any other background, the characters render tinted like this:
![image](https://user-images.githubusercontent.com/33850843/47958316-402b6e80-df86-11e8-9e31-ee4ec33a820f.png)

ryansloan:
I suspect this is an IE11 thing (possibly related to the use of `tint()`) Can you try each background effect (and refresh the page between them) and update this issue with which ones are broken?

ryansloan:
@boatmarina I suspect this is an IE11 thing (possibly related to the use of tint()) Can you try each background effect (and refresh the page between them) and update this issue with which ones are broken?

joshlory:
If the effect can instead tint the background with a changing color, and make the SVG a solid constant color (white?) I think that's the best fix.

PR #224, opened 2018-11-08
Fix IE 11 background effects

joshlory:
I was having trouble getting p5 to render the SVG in IE, so I used a tool to convert the SVG to canvas draw calls.  Also fixes the playground in IE 11.

Resolves https://github.com/code-dot-org/dance-party/issues/107.


src/Effects.js
```diff
@@ -472,3 +461,138 @@ module.exports = class Effects {
     };
   }
 };
+
+function drawSwirl(ctx) {
+  ctx.fillStyle = "#333";
```
islemaster: Should each of these be wrapped in `ctx.push()` and `ctx.pop()` calls to ensure we don't bleed drawing state into other operations?

islemaster: Never mind, I see this is part of the `draw` method above.

PR #229, opened 2018-11-09
Use karma for integration tests

cpirich:
* All tests under `test/integration` now run using `karma` with ChromeHeadless
* This adds about ~50 sec to our travis builds (3m20s vs 2m30s), but should allow us to do a much better job with testing in a "real" environment


joshlory:
This is awesome!!  Pulling down now to test it out.

webpack.config.js
```diff
@@ -1,4 +1,6 @@
 module.exports = {
+  // Uncomment when debugging
+  // devtool: 'eval-cheap-module-source-map',
```
islemaster: :joy: :sob:

joshlory: Better to set this automatically when `WATCH=1`?

cpirich: Yes, or maybe always for `yarn dev`? Seems like we should have source maps always? I wanted to check with you on this. But I've found myself needing it quite a bit to get work done, so I'm hesitant to have to keep stashing my change away.

.travis.yml
```diff
@@ -1,3 +1,8 @@
+dist: trusty
+
+addons:
+  chrome: stable
```
joshlory: Is the dist and addon change needed?  It looks like https://github.com/code-dot-org/craft/blob/master/.travis.yml is running ChromeHeadless without any added dependencies.

karma.conf.js
```diff
@@ -0,0 +1,55 @@
...
+    },
+    reporters: ['tap-pretty', 'coverage'],
+
+    // Run headless unless WATCH=1.
+    browsers: process.env['WATCH'] ? ['Chrome'] : ['ChromeHeadless', 'ChromeHeadlessNoSandbox'],
+
+    customLaunchers: {
+      ChromeHeadlessNoSandbox: {
+        base: 'ChromeHeadless',
+        flags: ['--no-sandbox']
```
joshlory: What is this needed for?

karma.conf.js
```diff
@@ -0,0 +1,55 @@
...
+    ],
+    preprocessors: {
+      'test/integration/*.js': ['webpack', 'sourcemap'],
+    },
+    webpack: {
+      node: {
+        fs: 'empty',
+      },
+      // Uncomment when debugging
+      // devtool: 'inline-source-map',
```
joshlory: Same comment as @islemaster's below.

test/integration/index.js
```diff
@@ -0,0 +1,3 @@
+require('./effectsLevelTest');
+require('./hourOfCodeTest');
+require('./spriteDanceTest');
```
joshlory: It looks like the tests aren't cleaning up after themselves, when running in the same context?  Probably non-blocking, but would be good to add a `teardown` step so only the currently-running visualization is visible.  We do something similar [in Craft integration tests](https://github.com/code-dot-org/craft/blob/b16dc2939cc30990c0134238f41b1d1c55664451/test/helpers/RunLevel.js#L38-L42).

screen shot 2018-11-09 at 12 22 12 pm

cpirich:
@joshlory. I want to change the test to reduce stubs. That is my follow up PR, but I don’t want to mix too many things together. This is a checkpoint to run what we were before, but in the new environment.

karma.conf.js
cpirich: This is from the Travis guide on how to use chrome headless in their container environment.

Same with the dist and addon change. (We may not need them, but without the addon we run on chrome 62 I believe). Just counting on whatever happens to be preinstalled

cpirich: This is from the Travis guide on how to use chrome headless in their container environment.

Same with the dist and addon change. (We may not need them, but without the addon we run on chrome 62 I believe). Just counting on whatever happens to be preinstalled

joshlory:
Makes sense!  Link to the next PR?

joshlory:
Two small nits when running `WATCH=1 npm run test:integration`.  The tests don't complete for me (I get the final line of output, but I have to manually kill the tests).  Also seeing this warning when the tests first start:
```
WARNING in configuration
The 'mode' option has not been set, webpack will fallback to 'production' for this value. Set 'mode' option to 'development' or 'production' to enable defaults for each environment.
```

.travis.yml
joshlory: Confirmed, it appears we don't need this: https://github.com/code-dot-org/dance-party/compare/use_karma_for_integration_tests...karma-next.

karma.conf.js
joshlory: (this shouldn't be needed if we specify webpack `mode: development`)

karma.conf.js
cpirich: Looks like it is not needed because we're not in the container based environment.

I was following this: https://docs.travis-ci.com/user/chrome

But then this is more important: https://blog.travis-ci.com/2018-10-04-combining-linux-infrastructures

test/integration/index.js
joshlory: Fixed in commit 6b44343082509995e0afdb1483cdbe0066bebf87.

PR #232, opened 2018-11-09
Fix move sorting again

cpirich:
* Even though we just fixed this sorting algorithm and added test cases, we were still susceptible to `Array.sort()` not being stable: see https://stackoverflow.com/q/3026281 (in particular Chrome <= 69 was unstable with arrays of more than 10 elements)
* Fixed by using 3 calls to `Array.filter()` which does preserve order: see https://stackoverflow.com/a/39712200

### DIFF
--- a/src/p5.dance.js
+++ b/src/p5.dance.js
@@ -110,7 +110,7 @@ module.exports = class DanceParty {
 
     // We can't use Array.sort() : see https://stackoverflow.com/q/3026281
     const restMoves = this.world.MOVE_NAMES.filter(move => move.rest);
-    const nonRestingFullLengthMoves = this.world.MOVE_NAMES.filter(move => !move.rest & !move.shortBurst);
+    const nonRestingFullLengthMoves = this.world.MOVE_NAMES.filter(move => !move.rest && !move.shortBurst);
     const shortBurstMoves = this.world.MOVE_NAMES.filter(move => move.shortBurst);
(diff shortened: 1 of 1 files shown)


cpirich:
Fixes https://github.com/code-dot-org/dance-party/issues/167 (again)

cpirich:
@islemaster not sure we have an obvious way to run regression tests (which we already wrote) on those specific browsers.
I'm having a separate discussion with Ryan on the use of move indices vs move names. I really wish we were using move names, but there may be some compatibility issues to consider.

joshlory:
Rather than relying on the order in the array + special indexes to denote rest vs. normal vs. burst, should we make the collection of dances nested one level deep, organized by rest/normal/burst?

islemaster:
What sort of compatibility issues?

joshlory:
I know it's urgent to get these fixes out quickly, but I wonder if we should come back and write this in a maintainable way.  The move ordering (as Brad mentioned) feels super fragile.

cpirich:
Yes, I've been thinking about the move index issue quite a bit. I think we should move to a nested approach or something like that.

I've been concerned that the student code that we persist (XML blocks) had indexes within, but it just refers to constants like `MOVES.ClapHigh` that we export from the interpreted code in this repo. So we may have a path to clean this all up without breaking existing projects.

PR #241, opened 2018-11-10
Combine spritesheets and json to minimize requests

islemaster:
**Before** we were downloading one png and one json metadata file for every animation for every character, making a total of 554 requests when loading the dance-party playground; not because we thought this was the _right_ thing to do, but because we were holding off on this optimization until our set of animations was fairly settled.

**After** we download one json file containing all the animation metadata, followed by ~five~ four spritesheets per character, making a total of ~60~ 49 requests when loading the dance-party playground. (Note - we still download almost exactly the same amount of data.)  I'm seeing local nocache playground load times consistently around 1200ms with the S3 cache configuration set properly and the cache warmed, where before they were closer to 2500.

A typical spritesheet is now ~2000x1200px, ~200k.  We're packing ~up to four long dances into a spritesheet, or up to seven short dances~ a handcrafted set of dances into each spritesheet.  See https://github.com/code-dot-org/dance-spritesheets for details on this process, which uses `mogrify` to pre-resize our frames, `TexturePacker` to arrange them, and `pngquant` to make them as small as possible.

The generated metadata is structured for ease of use, with all costume and move names normalized to lowercase:

```json
{
  "costume": {
    "movename": {
      "spritesheet": "basename of the corresponding png file",
      "frames": [
        {
          "frame": {},
          "sourceSize": {},
          "spriteSourceSize": {}
        }
      ]
    }
  }
}
```

This change depends on assets which I've already uploaded to a particular location on S3.

There is some risk of this change increasing the total memory footprint of Dance Party when all costumes are in use, due to inefficiencies in the packed textures.  I did some spot-checking and expect it's small, if at all.  I also expect that to be mitigated by the static analysis work Chris is doing.


islemaster:
@cpirich because I don't know how this will interact with the static analysis work you're doing.

@Hamms because this is yet another change to where we load sprites, and I don't know how much we need to change the server rendering to support this.

@breville for any further insight you have into how we pack and ship these sprites.  Do we group more aggressively and try to get down to 4 pngs per character?

islemaster:
I've learned that we were depending on having no move names to not try and load JSON during tests - working out proper stubbing for this now.

islemaster:
Okay, ready for review!  Tests fixed, and checked-in assets updated so integration tests are running against the new assets.  That means you'll want to skip to the end of the file list to see actual changes.

I've extracted a `ResourceLoader` class that allows us to easily point to the local asset base in integration tests, and inject a `UnitTestResourceLoader` for unit tests that provides stub images and metadata avoiding bad network dependencies.  @cpirich please let me know if this is going to mess with the work you're doing too much.

assets/sprite_sheets/characters.json
```diff
@@ -0,0 +1 @@
..."spriteSourceSize":{"x":114,"y":54,"w":74,"h":219}},{"frame":{"x":1584,"y":877,"w":82,"h":216},"sourceSize":{"w":300,"h":300},"spriteSourceSize":{"x":106,"y":57,"w":82,"h":216}},{"frame":{"x":1769,"y":668,"w":106,"h":215},"sourceSize":{"w":300,"h":300},"spriteSourceSize":{"x":95,"y":58,"w":106,"h":215}},{"frame":{"x":254,"y":882,"w":136,"h":212},"sourceSize":{"w":300,"h":300},"spriteSourceSize":{"x":79,"y":61,"w":136,"h":212}},{"frame":{"x":1459,"y":1,"w":151,"h":208},"sourceSize":{"w":300,"h":300},"spriteSourceSize":{"x":67,"y":65,"w":151,"h":208}},{"frame":{"x":830,"y":1,"w":157,"h":205},"sourceSize":{"w":300,"h":300},"spriteSourceSize":{"x":61,"y":68,"w":157,"h":205}},{"frame":{"x":671,"y":1,"w":157,"h":206},"sourceSize":{"w":300,"h":300},"spriteSourceSize":{"x":61,"y":67,"w":157,"h":206}},{"frame":{"x":989,"y":1,"w":156,"h":206},"sourceSize":{"w":300,"h":300},"spriteSourceSize":{"x":62,"y":67,"w":156,"h":206}},{"frame":{"x":1147,"y":1,"w":155,"h":207},"sourceSize":{"w":300,"h":300},"spriteSourceSize":{"x":63,"y":66,"w":155,"h":207}},{"frame":{"x":1304,"y":1,"w":153,"h":209},"sourceSize":{"w":300,"h":300},"spriteSourceSize":{"x":65,"y":64,"w":153,"h":209}},{"frame":{"x":1612,"y"
```
islemaster: This is the 'one true metadata' for all animations now. :grin:

breville:
Going from 5 to 4 spritesheets per character seems directionally good but likely offering diminishing returns.

The thing I wonder about the most is the earlier switch from canvas to Image, which presumably means that we're only storing PNG files in RAM most of the time now.  When we are rendering an individual sprite out of the PNG and onto a canvas, does the entire PNG need to be decoded to get the necessary pixels?

islemaster:
> ...does the entire PNG need to be decoded to get the necessary pixels?

I don't actually know.  Probably yes? Reading up a little bit on how PNG is encoded it seems like it'd be possible to decode only part of it, but I wouldn't depend on the browser using that capability.

> ... which presumably means that we're only storing PNG files in RAM most of the time now.

I don't think this is true.  My impression after a lot of profiling over the last couple of weeks is that Chrome decodes the entire PNG into video memory as a bitmap as soon as it's assigned to the `src` attribute for an image, even one not attached to the DOM, because decoding (while fast) is too slow to do during paint.

```js
const img = new Image();
img.src = 'spritesheet.png'; // Starts downloading _and_ decoding the PNG into vram
```

I might be wrong about that - I'm going to see if I can write a minimum repro to measure when Chrome decodes a PNG.

islemaster:
Whew... just spent an hour trying to learn how to read chrome://tracing and don't feel like I've gotten anywhere.  That's definitely where to look if we're trying to dig into lower-level performance stuff though.

src/ResourceLoader.js
```diff
@@ -0,0 +1,33 @@
...
+    this.assetBase_ = assetBase;
+  }
+
+  initWithP5(p5) {
+    this.p5_ = p5;
+  }
+
+  getAnimationData(callback) {
+    this.p5_.loadJSON(`${this.assetBase_}characters.json`, (json) => {
+      callback(json);
```
cpirich: can you just pass `callback` as the 2nd parameter?

src/p5.dance.js
```diff
@@ -191,35 +193,18 @@ module.exports = class DanceParty {
...
+    this.resourceLoader_.getAnimationData(animationData => {
+      this.world.SPRITE_NAMES.forEach(costume => {
+        const costumeData = animationData[costume.toLowerCase()];
+        ANIMATIONS[costume] = [];
+        this.world.MOVE_NAMES.forEach(({ name: moveName, mirror }, moveIndex) => {
+          const moveData = costumeData[moveName.toLowerCase()];
+          this.setAnimationSpriteSheet(
+            costume,
+            moveIndex,
+            this.resourceLoader_.loadSpriteSheet(moveData.spritesheet, moveData.frames),
```
cpirich: so we still call `loadSpriteSheet()` once per character per move, it's just that we refer to the same spritesheet URL in some places, so the "load" is instantaneous, correct? We will create multiple `Image` elements that refer to the same spritesheet, but hopefully the browsers are all smart enough to realize that they only need one underlying instance. If we were using `p5.Image` elements, I think this approach would be a little more dangerous. We may end up with multiple canvases with the exact same spritesheet in each.

Another approach (that may be required, but I'm not certain) would be to break the `Image` loading from the `SpriteSheet` loading. You'd first call our new `p5.loadImageElement()` with each unique URL (4-5 per character), then you can call `loadSpriteSheet()` (which would then no longer be async) for each dance, passing the loaded `Image` in as the 1st param (instead of the `url`)

islemaster:
Just updated with only four PNGs per character.  Updating description to match.

islemaster:
Whoa - there may be serious performance problems with this approach, when I run the following case:

```js
(function degenerateCase() {
  for (let cIndex = 0; cIndex < 11; cIndex++) {
    for (let move = 0; move < 12; move++) {
      const character = nativeAPI.world.SPRITE_NAMES[cIndex];
      const sprite = nativeAPI.makeNewDanceSprite(character, null, {
        x: 20 + 20 * move,
        y: 140 + 20 * cIndex,
      });
      nativeAPI.setProp(sprite, 'scale', 10);
      nativeAPI.changeMoveLR(sprite, move, 1);
    }
  }
}())
```

I made sure we're not creating multiple Image elements per asset, but I'm still seeing memory and framerate issues.  So I did a little profiling with chrome://tracing.

Here's a few seconds of very very busy GPU:

image

And here's 50ms of that very busy GPU:

image

And here's what's happening over and over:

image

That "FullImage" worries me.  Here are the details for one of those layers; you can see the "width" and "height" args are the size of a full spritesheet png.

image

Which seems like we're copying around entire spritesheets many times per frame, and I'm not sure why.

Curiously enough, I don't see this performance issue with this script that loads in the animations gradually, _until_ I switch tabs and come back.

```js
(function degenerateCase() {
  const characterCount = nativeAPI.world.SPRITE_NAMES.length;
  const timeBetweenAdds = 60;
  for (let cIndex = 0; cIndex < characterCount; cIndex++) {
    for (let move = 0; move < 12; move++) {
      const character = nativeAPI.world.SPRITE_NAMES[cIndex];

      setTimeout(() => {
        const sprite = nativeAPI.makeNewDanceSprite(character, null, {
          x: 30 + 30 * move,
          y: 30 + 30 * cIndex,
        });
        nativeAPI.setProp(sprite, 'scale', 15);
        nativeAPI.changeMoveLR(sprite, move, 1);
      }, cIndex * characterCount * timeBetweenAdds + move * timeBetweenAdds);
    }
  }
}());
```

Here's 50ms in the profiler after all the animations are loaded in for this case:

image

Night and day.  Here's the close-up:

image

It's copying a much smaller area.

image

When I switch away from the tab and come back, suddenly the profile looks like the first case.

I definitely feel like I'm lost down in Chrome-specific rendering behavior at this point.  The good news is this is probably not going to affect the cases we care about (fewer unique animations onscreen at once), and the memory footprint for the tab is staying under 150M.  Maybe we take a closer look at this on Monday?

islemaster:
Merging; I want to unblock Chris' work, and I my performance concerns here are limited to extreme cases that we're not officially supporting - definitely worth it in exchange for much better load times.

islemaster:
If we discover we need smaller spritesheets, this change still applies - instead of reverting it we'd just update our sprite packing configuration and push new assets.

PR #242, opened 2018-11-10
Measure meter

mrjoshida:
@ryansloan worked this up on the flight home to explore other ways to show measures. Wondering if this makes it easier to see what it means for an event to trigger "after" a given measure. Also experimenting with getting the measure visual out of the way of the dance.
measures


mrjoshida:
Attempting to make it easier to see exactly when one measure transitions to the next.
measures2

joshlory:
If we're worried about the risk of resizing the whole app, could we overlay the measure counter by the current "Measure: N" text?

PR #245, opened 2018-11-11
Add unit test for setBackground

madelynkasula:
Follow-up to #88


madelynkasula:
@joshlory no, those lock file changes aren't intentional -- must have come through in an `npm install` for some reason. the changes don't look harmful, but i'm going to remove them

PR #250, opened 2018-11-11
sparkles background effect

ryansloan:
Adds sparkles background effect based on Mark's mockups
floss


src/Effects.js
```diff
@@ -600,3 +633,56 @@ function drawSpiral(ctx) {
...
+  ctx.bezierCurveTo(25.2,30.7,23.6,29.1,0,27.2);
+  ctx.bezierCurveTo(23.6,25.2,25.2,23.6,27.2,0);
+  ctx.bezierCurveTo(29.1,23.6,30.7,25.2,54.3,27.2);
+  ctx.closePath();
+  ctx.fill();
+  ctx.stroke();
+  ctx.restore();
+  ctx.save();
+  ctx.fillStyle = color;
+  ctx.beginPath();
```
joshlory: Each of these sparkles actually has a second, hidden border path that I think we can skip drawing.  Fixing:

image

joshlory: Fixed in commit 32b81cd7566e7ac80416ab3b7477cd678228e437.

joshlory:
@Hamms PTAL!

PR #251, opened 2018-11-11
add new loader using dance party characters

simplycloud:
add a new p5 loading spinner with dance party characters during initial page load.

p5-loading


islemaster:
I pushed a [slightly different approach](https://github.com/code-dot-org/dance-party/pull/251/commits/f27f5c6ff67e65522332c2e6195be653aa42e6e7) to how this works for the demo page.  Key changes:

1. Instead of putting the loader markup in the document `<head>`, we actually put it inside the container p5 is about to take control of.  This makes sense for its purpose and helps with positioning/sizing.

2. The loader wasn't centered quite right.  Instead of using `position: fixed` which requires thorough knowledge of the page layout, we set the loader div to the known visualization size (400x400), depend on its parenting to position it correctly in the page, and then use flexbox (which [works great in our supported browsers](https://github.com/code-dot-org/dance-party/pull/251/commits/f27f5c6ff67e65522332c2e6195be653aa42e6e7)) to center the gif horizontally and vertically within the loader div.

islemaster:
Actual solution for milestone in https://github.com/code-dot-org/code-dot-org/pull/26002

PR #253, opened 2018-11-12
Enable static analysis with ensureSpritesAreLoaded

cpirich:
* We now defer loading spritesheets until either a new `ensureSpritesAreLoaded()` API is called (which takes an optional list of sprite names) or until `play()` is called. Both functions return a `Promise`. Additional API code should not be called the `Promise` from either of these functions is resolved. The expected calling pattern is to call `await ensureSpritesAreLoaded()` and then call `play()`.
* `dance-party` will now render "background only" while waiting for sprites to load.


package.json
```diff
@@ -44,6 +44,7 @@
     "raw-loader": "^0.5.1",
     "sinon": "^7.1.1",
     "tape": "^4.9.1",
+    "tape-async": "^2.3.0",
```
joshlory: Is this needed?  We already have tests that return promises that work as expected.

cpirich: Yes, I had no success without this. We can't successfully `await` and return in the test. The current `await` that we do for the API in unit tests is not a valid test case since the promise resolves immediately.

src/p5.dance.js
```diff
@@ -134,6 +138,68 @@ module.exports = class DanceParty {
...
+        // Already loaded, nothing to do:
+        return;
+      }
+      // We have found a new sprite to load, reset allSpritesLoaded to false
+      this.allSpritesLoaded = false;
+      const costumeData = animationData[costume.toLowerCase()];
+      this.world.MOVE_NAMES.forEach(({ name: moveName, mirror }, moveIndex) => {
+        const moveData = costumeData[moveName.toLowerCase()];
+
+        promises.push(new Promise(resolveSpriteSheet => {
```
islemaster: Cleanup idea: ResourceLoader's `loadSpriteSheet` can return a promise instead of taking a callback.

src/p5.dance.js
```diff
@@ -134,6 +138,68 @@ module.exports = class DanceParty {
...
+            }
+          );
+        }));
+      });
+    });
+    const promise = Promise.all(promises);
+    promise.then(() => {
+      this.allSpritesLoaded = true;
+    });
+    return promise;
```
islemaster: Not quite equivalent but maybe more idiomatic? (Ensures allSpritesLoaded gets set before anything chained outside this method is run)
```js
return Promise.all(promises).then(() => {
  this.allSpritesLoaded = true;
});
```

src/p5.dance.js
```diff
@@ -134,6 +138,68 @@ module.exports = class DanceParty {
...
+  }
+
+  async ensureSpritesAreLoaded(spriteNames) {
+    spriteNames = spriteNames || this.world.SPRITE_NAMES;
```
islemaster: Nit: Give the parameter a default.

src/p5.dance.js
```diff
@@ -247,7 +290,14 @@ module.exports = class DanceParty {
     }
   }
 
-  play(songData, callback) {
+  async play(songData, callback) {
```
islemaster: Seems like an antipattern for `play()` to both take a callback and return a promise.

src/p5.dance.js
```diff
@@ -247,7 +290,14 @@ module.exports = class DanceParty {
...
+    if (!this.lastEnsureSpritesPromise_) {
+      // The host has never called ensureSpritesAreLoaded(). We will call it ourselves here,
+      // effectively ensuring that all sprites are loaded.
+      this.ensureSpritesAreLoaded();
+    }
+    await this.lastEnsureSpritesPromise_;
```
islemaster: This also worries me.  We're sort-of guarding against multiple play calls while we are loading sprites, but in that case when the promise resolves don't we run the rest of `play()` multiple times back-to-back?  Seems like instead we should have a simple early-out if a previous play call has not completed.

src/p5.dance.js
cpirich: Either is fine with me. I just like having this function handle the toggling of that state variable.

src/p5.dance.js
cpirich: Sure, that would be a nice enhancement. As you know, this was written before ResourceLoader, so we were forced to adapt p5’s callbacks to promises. Now we have a layer in between.

src/p5.dance.js
cpirich: I’d like the default to be undefined. This way, older tests can still call it and get the full
set of moves. Would you prefer to see it as
`async ensureSpritesAreLoaded(spriteNames = undefined)` ?

islemaster: I'm saying this:
```js
  async ensureSpritesAreLoaded(spriteNames) {
    spriteNames = spriteNames || this.world.SPRITE_NAMES;
```
...could probably just be this?
```js
  async ensureSpritesAreLoaded(spriteNames = this.world.SPRITE_NAMES) {
```

src/p5.dance.js
cpirich: Agreed, but for now they complete at different times. This comes down to whether we care about allowing the dancers to appear before the song starts playing. Previously, that was important to some people. The promise fires immediately if you already have your sprite sheets ready. The callback is to signal the song starting.

And it is partially a case of migrating to the new contract. But, I’d also be ok with removing the callback and switching to a promise only. But the start of rendering will require that everything has loaded and the song is starting. So this is actually more of a change from the current behavior.

src/p5.dance.js
cpirich: This was intended to deal with the case of new sprites being needed (via a new call to `ensureSpritesAreLoaded()`) (for example, the student’s program changed and added a new character) - we don’t want the play() to start until that is complete.

We can change the contract to require that the caller called (and waited for `ensureSpritesAreLoaded()`). In which case, this isn’t needed and the play() call doesn’t need to return a promise. The CDO repo PR has made that the contract, so it’s just a matter of changing all of the tests in this repo.

src/p5.dance.js
cpirich: Updated with this change. 👍

src/p5.dance.js
cpirich: Updated with the 2nd approach, requiring `ensureSpritesAreLoaded()` to be called.

src/p5.dance.js
cpirich: Resolved by eliminating the `Promise` from `play()` - see below.

package.json
islemaster: I don't think this is needed - I just tore it out as an experiment, and all of your tests (including new ones) pass.

I'm pretty sure this works because `tape` is async-by-default and tests don't end when the function returns - they only end when [`test.end()`](https://github.com/substack/tape#tenderr) is called or the [`t.plan()`](https://github.com/substack/tape#tplann)ned number of assertions occur.

test/unit/whenKeyTest.js
```diff
@@ -2,9 +2,20 @@ const test = require('tape');
...
 test('make sure we are checking all relevant keys', async t => {
-  const nativeAPI = await helpers.createDanceAPI();
+  let lastEvents;
+  const nativeAPI = await helpers.createDanceAPI({ onHandleEvents: events => {
+    lastEvents = events;
+  }});
+
   const p5 = nativeAPI.p5_;
 
+  nativeAPI.play({
```
islemaster: Should we `await` this `nativeAPI.play()` call like you did elsewhere in tests?

islemaster:
I pushed several changes to simplify this diff, and am feeling fairly confident about this change.  I still feel a need to take another detailed look at the p5.dance.js changes before I approve - I'm worried about edge cases there.  Josh and Chris, this might be a good time to take another look.

test/unit/whenKeyTest.js
cpirich: No, `play()` no longer returns a `Promise` after your feedback. I probably need to remove the `await` in a few places now.

test/helpers/UnitTestResourceLoader.js
```diff
@@ -24,13 +24,7 @@ module.exports = class UnitTestResourceLoader {
...
-  loadSpriteSheet(baseName, frameData) {
-    return new Promise(resolve => {
-      const image = new Image();
-      const sheet = this.p5_.loadSpriteSheet(image, frameData);
-      setTimeout(() => {
-        resolve(sheet);
-      }, 0);
-    });
+  async loadSpriteSheet(baseName, frameData) {
+    return this.p5_.loadSpriteSheet(new Image(), frameData);
```
cpirich: @islemaster this change allows you to avoid `tape-async` - but we're also not really testing the asynchronous behavior anymore. Is that a good thing?

islemaster: Tests were passing without `tape-async` before I removed the `setTimeout` here (see Travis history on my commits).  I also suspect we are still genuinely testing async behavior because most promise implementations won't resolve synchronously.

Either way, in a perfect world promise resolution would be a black-box to our code - it shouldn't care whether a promise resolves synchronously or not.  I propose that it's nice to have coverage of both case (near-synchronous here in unit tests, and definitely async in our integration tests) to enforce that.

islemaster:
Cool, I feel a lot better about this after working with it for most of the day.  Just now I've been manually testing ed08239 in the dance-party playground and with https://github.com/code-dot-org/code-dot-org/pull/26001 and haven't found any issues.  I'm not seeing any errors or memory leaks across multiple runs, and it definitely cuts down on initial memory footprint and time-to-playable. (Unfortunately most of this delay now comes from code studio, not dance-party itself).

I'm still a little fuzzy on the effect this will have on the loading animation and click-to-play image, so I should probably do more verification around that.

I'm not yet set up to run thorough performance tests against this change, but could try to do that if we think it's the most useful next step.  @joshlory what are your thoughts on other verification we should do here before triage tomorrow?

test/helpers/UnitTestResourceLoader.js
cpirich: ok with me. several times while building this, I tried removing `tape-async` everywhere, but would see "text exited without end()" or something to that effect. But if it is passing now without it, we can leave it out.

islemaster: Ah - we actually had a test failure get into master branch a while back that looked like that, as a result of a test not cleaning up `sinon.useFakeTimers()`.  It's possible you were running into that.

islemaster:
We decided during triage today that we will not be accepting this change prior to Hour of Code.  We are still interested in taking it after Hour of Code.  We will revisit this on December 10th.

PR #367, opened 2018-11-14
add new backgrounds and foregrounds to dance party

simplycloud:
add confetti background to dance party

confetti


simplycloud:
going to close this and re-open a new one with new strategy for the foregrounds.

PR #373, opened 2018-11-14
Pass frame events to validationCallback

mrjoshida:
Levels 10 and 11 ask students to use keypress events, but there's no way for the validation callback to know if keypress events have fired. This passes `events` through so that we can validate againts actual fired events instead of just behavior that might suggest an event fired.


breville:
We've shipped this kind of validation before (e.g. starwars validated on all sorts of user actions during a level) so this seems good.

Issue #329, opened 2018-11-13
Backgrounds should paint fully before rendering each frame


breville:
The diamond background has a rendering issue in which the background isn't fully erased between frames:

![screenshot 2018-11-13 14 55 10](https://user-images.githubusercontent.com/2205926/48390053-4e327980-e754-11e8-8b8f-d9ee77748625.png)
![screenshot 2018-11-13 14 55 31](https://user-images.githubusercontent.com/2205926/48390054-4ecb1000-e754-11e8-90c9-b33cabfb7a01.png)

islemaster:
I've seen something similar with circles - wondering if we should just make sure we clear the whole frame before rendering any background.

PR #382, opened 2018-11-14
Ensure the canvas is cleared between each frame

joshlory:
Fixes https://github.com/code-dot-org/dance-party/issues/329.


breville:
@joshlory Any potential performance issue here?  Could this cause more pixels to be drawn than before?

PR #396, opened 2018-11-15
Kaleidoscope background effect

joshlory:
Add kaleidoscope background effect, with a few simple shapes.

kaleidoscope

### DIFF
--- a/src/Effects.js
+++ b/src/Effects.js
@@ -432,7 +432,9 @@ module.exports = class Effects {
 
     this.kaleidoscope = {
       init: function () {
-        this.shapes = p5.createGraphics(200, 200);
+        this.h = Math.sqrt(3) / 2 * 100;
+
+        this.shapes = p5.createGraphics(100, Math.ceil(this.h));
         this.shapes.pixelDensity(1);
         this.shapes.fill('white');
         this.shapes.noStroke();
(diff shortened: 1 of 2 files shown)


src/Effects.js
```diff
@@ -430,6 +430,113 @@ module.exports = class Effects {
...
+
+        const ctx = this.shapes._renderer.drawingContext;
+        ctx.save();
+        ctx.beginPath();
+        ctx.moveTo(50, 0);
+        ctx.lineTo(100, this.h);
+        ctx.lineTo(0, this.h);
+        ctx.clip();
+        this.shapes.clear();
+        this.shapes.rotate(p5.frameCount);
```
islemaster: Possibly unneeded optimization: It looks like the only thing that varies on the `shapes` buffer with each frame is its rotation, so I'm wondering if it's possible to do all of the drawing to `shapes` just once, and get the rotation effect with a frame-dependent transform in the `blitHex` operation.

src/Effects.js
```diff
@@ -430,6 +430,113 @@ module.exports = class Effects {
       }
     };
 
+    this.kaleidoscope = {
+      init: function () {
```
islemaster: Do we have any kind of `unload` for backgrounds?  Two 200x200 buffers (80kb) is not going to bust our memory budget, just wondering if in general we're able to clean up when the background is no longer in use.

src/Effects.js
joshlory: I like this.  I think I'll hold off on this for now, in case we want to make the shapes change with the music.  But if we see perf issues this would be a great optimization!

src/Effects.js
joshlory: Nothing yet, but I think we should look into this as part of the background crashes on real devices.

Issue #321, opened 2018-11-13
Can wreck browser with "every 0.001 measures"


breville:
It's possible to bring browser performance to its knees pretty easily:

![screenshot 2018-11-13 13 52 01](https://user-images.githubusercontent.com/2205926/48387812-576b1880-e74b-11e8-8c62-18aa051593ef.png)

ryansloan:
The youths do like to find ways to crash code.org. Can we just floor it at some reasonable place? Not sure where the reasonable line is, but from a product perspective we need to support .25 because quarter notes are an interesting unit that we've seen people explore. I am fine drawing the line at .25 unless @mrjoshida has a better recommendation

islemaster:
I did a little initial exploration of this and was going to suggest 0.05 as a minimum value for both measures and seconds.  That's 20 events per second (less than once per frame) and allows sixteenth notes both when given in measures (0.0625 measures) or in seconds at most tempos (0.05 seconds up to 240bpm).

If we're worried that will still crush performance, bumping the minimum to 0.1 (quarter) notes puts us at a maximum of 10 events per second, which is pretty typical for an event loop in a game.

ryansloan:
@islemaster Let's use .1

joshlory:
(We'll want to restrict this in the block, and also validate it at the API layer.)

islemaster:
Fix shipped yesterday!

PR #409, opened 2018-11-15
Maximum event rate

islemaster:
Partial fix for https://github.com/code-dot-org/dance-party/issues/321

Ensures we don't fire events more than once every 100ms when a student writes code like `every 0.001 measures` or `every 0.001 seconds`.

@ryansloan requested a limit of 0.1 for both seconds and measures, which effectively limits us to events on eighth notes on our fastest song at 169bpm.

The other half of this work will happen in the code-dot-org repo, enforcing the limit at the block UI layer.

### DIFF
--- a/src/p5.dance.interpreted.js
+++ b/src/p5.dance.interpreted.js
@@ -228,6 +228,20 @@ function everySecondsRange(n, unit, start, stop, func) {
     return;
   }
 
+  // Limit minimum event interval to avoid performance problems
+  // from student code like `every 0.001 seconds do...`
+  if ('seconds' === unit) {
+    // Maximum of 10 events per second
+    // Our fastest song is 169bpm
+    // This allows events on eighth-notes in that song.
+    n = Math.max(0.1, n);
+  } else {
+    // Maximum of ten events per measure
+    // Our fastest song is 169bpm
+    // This allows events every 142ms in that song
+    n = Math.max(0.1, n);
+  }
+
   // Offset by n so that we don't generate an event at the beginning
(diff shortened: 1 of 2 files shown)


src/p5.dance.interpreted.js
```diff
@@ -228,6 +228,20 @@ function everySecondsRange(n, unit, start, stop, func) {
     return;
   }
 
+  // Limit minimum event interval to avoid performance problems
+  // from student code like `every 0.001 seconds do...`
+  if ('seconds' === unit) {
```
breville: ordering_meme

test/unit/eventsTest.js
```diff
@@ -142,3 +142,41 @@ test('non-conflicting atTimestamp cues, both take effect', async t => {
...
   nativeAPI.reset();
 });
+
+test('silently clamp maximum event rate in seconds', async t => {
+  // Try and create an event that's waaay too frequent
+  const {nativeAPI, interpretedAPI} = await runUserCode(`
+    everySeconds(0.00314, "seconds", function () {});
+  `);
+
+  const assertClose = (a, b) => t.ok(Math.abs(a - b) < 1.0e-7, `${a} and ${b} were within 1.0e-7`);
```
breville: Excellent.

src/p5.dance.interpreted.js
islemaster: [Yoda conditions](https://en.wikipedia.org/wiki/Yoda_conditions) forever!

PR #422, opened 2018-11-17
V1 for new fireworks effect.

raytrodd:
fireworks

### DIFF
--- a/src/Effects.js
+++ b/src/Effects.js
@@ -600,8 +600,132 @@ module.exports = class Effects {
         });
       }
     };
+
+    this.fireworks = {
+      fireworks:[],
+
+      makeRandomFirework: function () {
+        var ret =  {
+          rocket:null,
+          particles:[],
+          exploded:false,
+
+          update: function() {
+            this.rocket.update()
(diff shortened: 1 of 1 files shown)


ryansloan:
With tracers looks *awesome*, but can we test on some lower-end devices? seems like a lot of particles to keep track of

src/Effects.js
```diff
@@ -600,9 +600,153 @@ module.exports = class Effects {
...
+        };
+      },
+
+      draw: function () {
+        if (this.buffer === null) {
+          this.buffer = p5;
+          // We get the tracer effect by writing frames to a
+          // offscreen buffer that has a transparent background
+          if (this.enableTracers) {
+            this.buffer = p5.createGraphics(p5.width, p5.height);
```
islemaster: @cpirich identified a memory leak in Safari when using p5.createGraphics in a background effect, documented in https://github.com/code-dot-org/dance-party/pull/26.  Can we review carefully for that?

simplycloud: agreed to merge and test performance in testing / staging.

PR #424, opened 2018-11-18
v1 add new bubbles and confetti foregrounds

simplycloud:
add new bubbles and confetti foregrounds to dance party. this is a v1 and we decide if we want to randomize the colors to add more variety.

confetti

bubbles


simplycloud:
going to break up the two foregrounds and look at why confetti foreground is failing CI tests.

PR #425, opened 2018-11-18
add bubbles foreground to dance party

simplycloud:
add bubbles foreground to dance party. can update color scheme based on feedback.

bubbles

### DIFF
--- a/src/Effects.js
+++ b/src/Effects.js
@@ -600,6 +600,36 @@ module.exports = class Effects {
         });
       }
     };
+
+    this.bubbles = {
+      bubble: [],
+      draw: function () {
+        let bubble = {
+          x: p5.random(-100, 400),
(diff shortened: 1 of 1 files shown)


src/Effects.js
```diff
@@ -600,6 +600,36 @@ module.exports = class Effects {
...
+
+    this.bubbles = {
+      bubble: [],
+      draw: function () {
+        let bubble = {
+          x: p5.random(-100, 400),
+          y: 410,
+          velocityX: p5.random(-2, 2),
+          size: p5.random(6, 12, 18),
+          color: randomColor(100, 50, 0.25),
```
ryansloan: I seem to remember there were some issues (memory, I think) with rendering things with transparencies - @joshlory or @islemaster do you have any memory of this?

PR #426, opened 2018-11-18
v1 add strobe lights background to dance party

simplycloud:
add strobe lights background to dance party. i think this looks good but i'm seeing some slight memory leaks demonstrated by light flashing and character slowdown after about 30 seconds. would like to see if we can optimize this a bit further.

strobe_lights

### DIFF
--- a/src/Effects.js
+++ b/src/Effects.js
@@ -600,6 +600,34 @@ module.exports = class Effects {
         });
       }
     };
+
+    this.strobe_lights = {
+      strobe: [],
+      draw: function () {
+        p5.background('black');
+        p5.strokeWeight(5);
+        p5.stroke(255);
+        let strobe = {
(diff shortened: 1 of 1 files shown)


src/Effects.js
```diff
@@ -600,6 +600,34 @@ module.exports = class Effects {
...
+        p5.strokeWeight(5);
+        p5.stroke(255);
+        let strobe = {
+          lineH: 25,
+          lineW: 1,
+          lineX: 0,
+          lightH: 50,
+          lightD: 40,
+        };
+        this.strobe.push(strobe);
```
joshlory: It looks like we never remove strobes from the list?

simplycloud:
closing and addressing in https://github.com/code-dot-org/dance-party/pull/448

PR #427, opened 2018-11-18
update tacos foreground to use SVG

simplycloud:
update tacos foreground to use SVG. ran optimizations on SVG using SVGOMG tool.

tacos


simplycloud:
@ryansloan should we see about making the tacos larger to match the v1 tacos foreground?

simplycloud:
closing and opening new PR.

PR #430, opened 2018-11-18
add hearts foreground effect

simplycloud:
add hearts foreground effect to dance party.

hearts


simplycloud:
closing and opening a new PR.

PR #434, opened 2018-11-19
v1 add floating stars background

simplycloud:
add floating stars background to dance party. will take feedback and suggestions if there is a way to make this look better.

bear-stars


simplycloud:
closing and addressing in https://github.com/code-dot-org/dance-party/pull/448

PR #419, opened 2018-11-16
Background Visual Tests

epeach:
This is a first pass - I'd love any feedback on strategy or implementation that you have - thanks!
This PR:
* Sets up visual testing for dance-party
* Includes test that generates screenshots and compares local screenshots to accepted screenshots for all backgrounds
* Includes accepted screenshots
* Adds an option to DanceParty to turn off randomness - currently only used in Effects.js
* Updates snowflakes effect to use internal randomNumber generator (where randomness can be turned off)

PR #441, opened 2018-11-19
Fix visual test and snowflakes

epeach:
While neatening up the visual test code here (https://github.com/code-dot-org/dance-party/pull/419) I moved away from the suggested code from pixelmatch.

This broke the tests and made them always pass. For the sake of getting this fixed ASAP, using the pixelmatch suggested code again.

Also, because the tests always passed, it didn't fail when I reverted the randomnumber in snowflakes to use p5.random. Once I fixed the tests, this test failed (due to the randomness change). Adding an updated snowflakes to match the new accepted background.


test/visual/backgrounds.js
```diff
@@ -22,15 +20,19 @@ async function createScreenshot(effectName) {
 async function testBackground(t, effect) {
   await createScreenshot(effect);
 
-  const [actual, expected] = await Promise.all([
-    readPNG(`${tempDir}/${effect}.png`),
-    readPNG(`${fixturePath}${effect}.png`)
-  ]);
```
islemaster: I see what went wrong here.  `readPNG` wasn't resolving to a value at all.  It's interesting that this failed quietly.

If you wanted to stick with the `Promise.all` approach I _think_ you'd just need to fix `readPNG`, which would look something like this:

```js
function readPNG(pngPath) {
  return new Promise(resolve => {
    const stream = fs.createReadStream(pngPath)
      .pipe(new PNG())
      .on('parsed', () => resolve(stream));
  });
}
```

islemaster:
FWIW I vote we merge this out-of-band since it's a test fix.

epeach:
I'll implement the suggested change and try to figure out why raining_tacos is still failing, and after that I'll touch base with you about merging out-of-band

epeach:
@islemaster and @joshlory - any ideas why this might be failing on raining_tacos? I'm not able to repro locally and the accepted baseline on master looks like my local.

test/visual/backgrounds.js
```diff
@@ -49,7 +49,7 @@ async function testBackground(t, effect) {
   'strobe',
   'rain',
   'text',
-  'raining_tacos',
+  //'raining_tacos', temporarily disable: to figure out emoji compatibility
```
epeach: Per discussion with @islemaster to unblock the rest of these changes

epeach:
Merged outside of triage with Dotd approval

PR #442, opened 2018-11-19
sprites make relative changes for most properties

Bjvanminnen:
This is one part of https://github.com/code-dot-org/dance-party/issues/75

It makes it so that when we use the block that makes a sprite follow bass/treble/mid, we do so relative to the current state (except for tint).

For x, y we map to a range from (current - 150, current + 150)
For scale, width, height, we map from (current * 0.5, current * 1.5)
For rotation we map to a range (current - 90, current + 90).

I've removed the unused "direction" property.

### DIFF
--- a/src/p5.dance.js
+++ b/src/p5.dance.js
@@ -854,21 +854,24 @@ module.exports = class DanceParty {
    * @param {string} range
    */
   startMapping(sprite, property, range) {
+    if (!sprite) {
+      return;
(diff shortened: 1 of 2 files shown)


joshlory:
It looks like the base rotation is 90 degrees to the left?  (Ah, I see that's how it's spec'd — do we want it to be 0 to 180?)

image

PR #447, opened 2018-11-20
Improve depth sorting

breville:
This makes two proposed improvements to depth sorting:

- Scale becomes the most important factor.  The bigger the scale, the higher in the stack the sprite is drawn.  Y coordinate becomes the second tie-breaker, and X coordinate becomes the third.

- We now adjust the depth of every sprite after any operation that adjusts scale, x, or y.  Previously, depth was only set for sprites in a group layout and only at the time of setting that layout.

Here are a couple examples:

In a group layout, fewer sprites in a group means a bigger scale, and now that means those bigger sprites end up in front:

screenshot 2018-11-20 18 48 46

Now, even after periodically adding regular sprites, depth is applied.  For sprites that are the same scale, Y coordinate becomes the tie-breaker:

screenshot 2018-11-20 18 46 23

### DIFF
--- a/src/p5.dance.js
+++ b/src/p5.dance.js
@@ -371,10 +371,15 @@ module.exports = class DanceParty {
         sprite.x = position.x;
         sprite.y = position.y;
       }
+      this.adjustSpriteDepth_(sprite);
     };
     sprite.setScale = function (scale) {
       sprite.scale = scale;
+      this.adjustSpriteDepth_(sprite);
     };
+
+    this.adjustSpriteDepth_(sprite);
+
     return sprite;
   }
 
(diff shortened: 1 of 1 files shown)


ryansloan:
@breville, this is great. The use of size as a factor is (I think) just the rule we were looking for when trying to improve the coordinate-based Z-ordering. Thank you for wrangling this!

src/p5.dance.js
```diff
@@ -826,6 +837,16 @@ module.exports = class DanceParty {
...
   }
 
+  adjustSpriteDepth_(sprite) {
+    if (!this.spriteExists_(sprite)) {
+      return;
+    }
+
+    // Bias scale heavily (especially since it largely hovers around 1.0) but use
+    // Y coordinate as the first tie-breaker and X coordinate as the second.
+    sprite.depth = sprite.scale * 100 + sprite.y / 400 + sprite.x / 40000;
```
islemaster: It took me a minute to understand why 400 and 40000 here.  I wonder if we want to separate the weight and normalization for readability.
```js
sprite.depth =
  10000 * sprite.scale +
  100 * sprite.y / 400 +
  1 * sprite.x / 400; 
```

breville: Nice!

joshlory:
Also a good candidate for visual tests!  (post-merge)

ryansloan:
Before we ship this, we want to look at the way that the new depth-sorting logic works in conjunction with the "following" block today.

PR #448, opened 2018-11-20
addition and cleanup of background and foreground scenes

simplycloud:
addition and cleanup of background and foreground scenes. test each on safari, firefox, and chrome.

- svg cleanup for taco, music notes, falling poop, pizzas, pineapple, hearts, and rainbows
- added starspace background
- added strobelights background
- added stars backgound
- added rainbows and hearts foreground
- updated confetti to have the confetti spin
- updated music_notes to be responsive to music
- removed original strobe background

note: will clean up one change in next PR for strobelights

### DIFF
--- a/src/Effects.js
+++ b/src/Effects.js
@@ -129,29 +129,6 @@ module.exports = class Effects {
       }
     };
 
-    this.strobe = {
-      waitTime: 0,
-      flashing: false,
-      update: function () {
-        this.flashing = true;
-        this.waitTime = 6;
-      },
-      draw: function ({isPeak}) {
-        let bgcolor = p5.rgb(1, 1, 1);
-        if (isPeak) {
-          this.update();
-        }
-        if (this.flashing) {
-          this.waitTime--;
-        } else {
-          p5.background(bgcolor);
-        }
-        if (this.waitTime <= 0) {
-          this.flashing = false;
-        }
-      }
-    };
-
     this.rain = {
       drops: [],
       init: function () {
@@ -185,6 +162,7 @@ module.exports = class Effects {
         p5.pop();
       }
     };
+
     this.sparkles = {
       sparkles: [],
       maxSparkles: 80,
@@ -219,6 +197,7 @@ module.exports = class Effects {
         }
       },
     };
+
     this.text = {
       texts: [],
       maxTexts: 10,
@@ -281,9 +260,9 @@ module.exports = class Effects {
           const taco = this.tacos[i];
           p5.translate(taco.x, taco.y);
           p5.rotate(taco.rot);
-          p5.textAlign(p5.CENTER, p5.CENTER);
-          p5.textSize(this.size);
-          p5.text(String.fromCodePoint(55356, 57134), 0, 0);
+          p5.scale(2, 2);
+          drawTaco(p5._renderer.drawingContext);
+          p5.pop();
           taco.y += taco.speed;
           taco.rot++;
           if (taco.y > 450) {
@@ -320,7 +299,9 @@ module.exports = class Effects {
           const pineapple = this.pineapple[i];
           p5.translate(pineapple.x, pineapple.y);
           p5.rotate(pineapple.rot);
+          p5.scale(2, 2);
           drawPineapple(p5._renderer.drawingContext);
+          p5.pop();
           pineapple.y += pineapple.speed;
           pineapple.rot++;
           if (pineapple.y > 450) {
@@ -357,7 +338,9 @@ module.exports = class Effects {
           const pizza = this.pizza[i];
           p5.translate(pizza.x, pizza.y);
           p5.rotate(pizza.rot);
+          p5.scale(2, 2);
           drawPizza(p5._renderer.drawingContext);
+          p5.pop();
           pizza.y += pizza.speed;
           pizza.rot++;
           if (pizza.y > 450) {
@@ -406,6 +389,7 @@ module.exports = class Effects {
         }
       }
     };
+
     this.swirl = {
       angle: 0,
       color: null,
@@ -428,6 +412,7 @@ module.exports = class Effects {
 
       }
     };
+
     this.spiral = {
       angle: 0,
       color: null,
@@ -487,7 +472,6 @@ module.exports = class Effects {
         if (this.targetX === 0) {
           this.init();
         }
-
         p5.push();
         p5.noFill();
         p5.strokeWeight(600);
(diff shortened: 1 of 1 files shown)


src/Effects.js
```diff
@@ -281,9 +260,9 @@ module.exports = class Effects {
           const taco = this.tacos[i];
           p5.translate(taco.x, taco.y);
           p5.rotate(taco.rot);
-          p5.textAlign(p5.CENTER, p5.CENTER);
-          p5.textSize(this.size);
-          p5.text(String.fromCodePoint(55356, 57134), 0, 0);
+          p5.scale(2, 2);
+          drawTaco(p5._renderer.drawingContext);
+          p5.pop();
```
islemaster: FYI @epeach soon the taco will not be an emoji and we can re-enable it in visual tests!

src/Effects.js
```diff
@@ -357,7 +338,9 @@ module.exports = class Effects {
           const pizza = this.pizza[i];
           p5.translate(pizza.x, pizza.y);
           p5.rotate(pizza.rot);
+          p5.scale(2, 2);
           drawPizza(p5._renderer.drawingContext);
+          p5.pop();
```
islemaster: We've now got one `push()` and two `pop()` calls in each of these draw methods, which seems like a bug.  Can you explain why `pop()` is needed here when it also exists below (line 350)?

src/Effects.js
```diff
@@ -670,7 +655,10 @@ module.exports = class Effects {
           const poop = this.poop[i];
           p5.translate(poop.x, poop.y);
           p5.rotate(poop.rot);
+          p5.push();
+          p5.scale(2, 2);
           drawFallingPoop(p5._renderer.drawingContext);
+          p5.pop();
```
islemaster: This one is balanced, but I don't think we need this `push/pop` at all - no drawing operations occur between this `pop()` and the next one, suggesting they're redundant.

src/Effects.js
```diff
@@ -682,6 +670,85 @@ module.exports = class Effects {
...
+    this.hearts = {
+      heart: [],
+      size: 50,
+      init: function () {
+        for (let i = 0; i < 20; i++) {
+          this.heart.push({
+            x: randomNumber(10, 390),
+            y: -50,
+            rot: randomNumber(0, 359),
+            speed: p5.random(1,5),
```
islemaster: I missed this somehow - why are we using `randomNumber` in some cases and `p5.random` in others?

src/Effects.js
```diff
@@ -682,6 +670,85 @@ module.exports = class Effects {
...
+      update: function () {
+        this.size += randomNumber(-5, 5);
+      },
+      draw: function () {
+        if (this.heart.length < 1) {
+          this.init();
+        }
+        for (let i = 0; i < this.heart.length; i++) {
+          p5.push();
+          const heart = this.heart[i];
```
islemaster: Nit: I find this naming confusing.  I'd prefer `this.hearts` or `this.heartList` for the array.

src/Effects.js
```diff
@@ -743,6 +810,98 @@ module.exports = class Effects {
...
+            p5.ellipse(0, 0, 1, star.size);
+          }
+          let fadeSpeed = p5.map(star.size, 15, 30, 1, 2);
+          star.size = star.size - fadeSpeed;
+          star.y = star.y - 2;
+          p5.pop();
+        });
+        this.star = this.star.filter(function (star) {
+          return star.size > 0.1;
+        });
```
islemaster: If I'm reading this correctly, your shortest-lived stars last 73 frames, and your longest-lived stars last 83 frames, so we should expect to be drawing about 78 stars on any given frame.  Does that sound right?

I wonder if we should pass the optional `withinBounds` flag to [`p5.map`](https://p5js.org/reference/#/p5/map) so that the lowest fadeSpeed is 1.

src/Effects.js
```diff
@@ -743,6 +810,98 @@ module.exports = class Effects {
...
+        }
+        p5.noStroke();
+        p5.fill('white');
+        this.space.forEach(function (space){
+          p5.push();
+          p5.translate(space.x, space.y);
+          p5.ellipse(0, 0, space.size, space.size);
+          let speedMultiplier = p5.pow(space.size, 2) /2;
+          space.x += space.velocity.x * speedMultiplier;
+          space.y += space.velocity.y * speedMultiplier;
```
islemaster: Sooo... velocity is a constant, and acceleration is controlled by the radius of the ellipse?

src/Effects.js
simplycloud: will address these throughout. thanks!

src/Effects.js
simplycloud: will address these throughout. thanks!

src/Effects.js
simplycloud: as discussed, probably just carry over from re-using other effects that chose different styles. will look to standardize across the file as part of a clean up PR later.

src/Effects.js
```diff
@@ -1018,198 +1149,226 @@ function drawFallingPoop(ctx) {
...
-  ctx.restore();
-  ctx.save();
-  ctx.fillStyle = "#995243";
-  ctx.beginPath();
-  ctx.moveTo(9.21,2.183);
-  ctx.bezierCurveTo(6.843,2.603,5.45,0,5.45,0);
-  ctx.bezierCurveTo(4.062,2.411,4.166,5.546,6.356,7.264);
-  ctx.bezierCurveTo(7.734,8.345,9.317,9.11,10.58,8.406);
-  ctx.translate(10.29000831352916,5.243266874714561);
+  ctx.moveTo(5.449,0);
```
breville: What is the source of all the SVG code in this file?

src/Effects.js
simplycloud: fixed. used heartList.

src/Effects.js
simplycloud: discussed improving in next version.

src/Effects.js
simplycloud: discussed looking further at this in next version.

src/Effects.js
simplycloud: mark made these SVG drawings for us.

breville: Can we document the pipeline used here so that future maintainers know how to work with it?

I was worried this was hand-coded :)

It's also good to know whether we can round-trip the assets (from code back into tool), but it looks like it's one-directional here, which means knowing where they came from is quite important.

joshlory:
image

😱

Issue #199, opened 2018-11-08
Half of the dance actions goes in the opposite direction than indicated in the "do once" blocks


poorvasingal:
In half of the dance moves, if you choose the "do once" block and choose the left direction, the move will actually happen to the right (and vice versa). Note that the direction seems to be correct if using the "do forever" blocks.

Incorrect dance moves: Clown, Double Jam, Floss (I think), and Roll.

uponthesun:
Dug into this a little as a learning exercise, and as far as I can tell, this is because of this logic which sets the animation frame to halfway through the animation:

https://github.com/code-dot-org/dance-party/blob/master/src/p5.dance.js#L472-L473

I tried changing the code locally to set the frame number to 0 instead, and it uses the correct direction then.

I imagine just reversing the mirror direction would "fix" this, though that doesn't seem great... hopefully there's a better way?

islemaster:
Testing in the playground with this code
```js
makeNewDanceSprite("CAT", null, {x: 200, y: 200});

var move = 0;
everySeconds(0.5, "measures", function () {
  doMoveEachLR('all', move, 1);
  move = (move + 1) % 25;
});
```

islemaster:
Fix is scheduled for the next deploy.

PR #449, opened 2018-11-21
Mirror certain burst moves

islemaster:
Fixes https://github.com/code-dot-org/dance-party/issues/199

Causes certain burst moves to be mirrored according to what feels most natural for a particular move.  We play burst moves starting from the middle of longer animations, and for some of them that involve back-and-forth movement that middle point looks like it's going the wrong direction.

Easier to show than tell.  Here's all of the burst moves playing "to the right", before this change:

before_totheright

And here are all of them playing "to the right" after this change:

after_totheright

The ones I've flipped are Clown, DoubleJam, Floss, Kick, Roll, ThisOrThat, Thriller, and XHeadHips.

Note that, in a lot of ways, this is going to come down to "feel" and I doubt we'll get everyone to agree on which direction is best to associate with some of these.


joshlory:
@islemaster can we add visual tests for this (after merge)?

PR #450, opened 2018-11-21
use a set of representative measures to modify energy

Bjvanminnen:
A second part of https://github.com/code-dot-org/dance-party/issues/75

The other piece is https://github.com/code-dot-org/dance-party/pull/442 (making movement relative instead of absolute).

This piece makes it so that when calculating the mean/standard deviation of the energy for a song, we do so across a representative set of measures (5 to 12 by default) instead of across the entire song.  We believe this range to be effective for most songs, but may want to end up tuning it for a few manually.

It is likely that we could accomplish the same sort of thing by customizing the numDeviations options on a song by song basis, but this was decided that it's easier for the human tuner to reason about representative measures.

This code all better belongs in the preRender step, but that would mean regenerating all of our song metadatas. We do think there will still be some metadata regenerated for songs where the default representative measures are no good.

The main risk here is that we make the energy levels worse for some songs.

### DIFF
--- a/src/modifySongData.js
+++ b/src/modifySongData.js
@@ -1,3 +1,10 @@
+// This is derived from DanceParty.getCurrentMeasure, but don't want to explicitly
+// depend on DanceParty so that this could be easily moved to a preRender step in
+// the future.
+function getMeasureForTime(songMetadata, time) {
+  return songMetadata.bpm * ((time - songMetadata.delay) / 240) + 1;
+}
+
 function clamp(min, max, val) {
   return Math.max(
     min,
@@ -10,6 +17,13 @@ function calculateMean(vals) {
   return sum / vals.length;
 }
 
+function valueOrDefault(value, defaultValue) {
+  if (value === undefined) {
+    return defaultValue;
+  }
+  return value;
+}
+
 /**
  * This method takes an input list of energy values, and outputs a new list. THe
  * methodology for figuring out the new list is to calculate the mean and standard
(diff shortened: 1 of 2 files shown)


Bjvanminnen:
I've tested every song to make sure the songMetada gets processed without error. I also added some perf logging, and the modifySongMetaData step appears to take on the order of < 10ms

src/modifySongData.js
```diff
@@ -1,3 +1,10 @@
+// This is derived from DanceParty.getCurrentMeasure, but don't want to explicitly
+// depend on DanceParty so that this could be easily moved to a preRender step in
+// the future.
+function getMeasureForTime(songMetadata, time) {
+  return songMetadata.bpm * ((time - songMetadata.delay) / 240) + 1;
```
islemaster: I find the redundant parens actually make it harder to figure out why this math is the way it is... it took me a while to work out why we're using `240` here.  Might be worth this for readability:
```js
const measuresPerSecond = songMetadata.bpm / 240;
const secondsElapsed = time - songMetadata.delay;
return secondsElapsed * measuresPerSecond + 1;
```
I'd also document that the `time` parameter here is in seconds, not milliseconds, since it's less idiomatic for JavaScript.

src/modifySongData.js
```diff
@@ -19,13 +33,21 @@ function calculateMean(vals) {
...
-function getFrequencyEnergy(energy, numDeviations=1.5, smoothFactor=0.7) {
-  const nonZero = energy.filter(e => e > 0);
+function getFrequencyEnergy(energy, options = {}) {
+  const numDeviations = valueOrDefault(options.numDeviations, 1.5);
+  const smoothFactor = valueOrDefault(options.smoothFactor, 0.7);
+  const representativeIndexRange = valueOrDefault(options.representativeIndexRange ||
+    [0, energy.length]);
+
+  const nonZero = energy.slice(...representativeIndexRange).filter(e => e > 0);
   const mean = calculateMean(nonZero);
```
islemaster: Nit: Do you really mean `nonZero` here or `greaterThanZero`, or is it irrelevant because we never see negative values?

test/unit/modifySongDataTest.js
```diff
@@ -2,10 +2,15 @@ const test = require('tape');
 const modifySongData = require('../../src/modifySongData');
 const songData = require('../../metadata/jazzy_beats.json');
 
+function songDataWithCustomOpts(energyCustomization) {
+  return Object.assign({}, songData, { energyCustomization });
```
islemaster: Nonblocking, just curious: Do we not support object-spread in this repo yet?  Or is there a reason you prefer `Object.assign`?
```js
return {...songData, energyCustomization};
```

src/modifySongData.js
Bjvanminnen: It is irrelevant here because values are all positive.

test/unit/modifySongDataTest.js
Bjvanminnen: That is correct. It seemed to work in the test (at least when I ran via `node`), but `...` did not work in the main code base, so I decided to be consistent about using `assign`

PR #455, opened 2018-11-21
Backgrounds Visual Tests - Improvement and Additions

epeach:
While adding test coverage for the new backgrounds, I made the following improvements to the background visual tests.

Tests Added: 
raining_tacos (re-enabled after moving away from emojis)
snowflakes, sparkles, pineapples, pizzas, falling_poop, hearts, falling_rainbows, bubbles, stars, starspace, strobe_lights, confetti, music_notes

Test Framework Improvements:
*Provides text for text background
*Every 6 draws, isPeak is true -> this triggers things like color changes and coordinate changes in the backgrounds so the tests more accurately represent what the user sees
*Provides a centroid - effects like music_notes rely on this value. Currently provides a hard-coded number in the expected range.
*Clears background between 'draws' -> now only see current location of effects like tacos and spotlight

Note: Kaleidoscope is currently disabled while I investigate an inconsistency between local and Travis screenshots.


src/Effects.js
```diff
@@ -654,7 +654,6 @@ module.exports = class Effects {
           p5.rotate(poop.rot);
           p5.scale(2, 2);
           drawFallingPoop(p5._renderer.drawingContext);
-          p5.pop();
```
epeach: @simplycloud It looks like there were two calls to `p5.pop`. I removed one, but want to double check with you that this won't break expected outcomes.

epeach:
I'm still investigating why pineapple and pizza only generate one image (and in pizza's case, only a partial). I'm in the process of narrowing down why that is the case - so far, I can confirm that the correct number of images, each with unique x and y, are being generated and their values appropriately updated. I'm currently investigating the drawing process to see if I can identify why they appear to be drawn on top of each other. I'll comment out the pineapple and pizza tests and remove the 'accepted' baselines from this PR with the intention that they'll be re-added once they are drawing more than one image or that we have an understanding of why they are drawn on top of each other.

Issue #459, opened 2018-11-22
Measure counter renders wrong in RTL


breville:
This is what 10 looks like:

![screenshot 2018-11-22 11 27 29](https://user-images.githubusercontent.com/2205926/48874864-adc21080-ee49-11e8-9db0-f7a35c328537.png)

breville:
Also am not sure why ":Select song" has the colon on the left.

Also am not sure why the instructions have a period on the left:

![screenshot 2018-11-22 11 34 24](https://user-images.githubusercontent.com/2205926/48875039-a2231980-ee4a-11e8-8c2a-df56dc270379.png)

breville:
From discussion with @hamms, the fix in https://github.com/code-dot-org/dance-party/pull/461 is good enough for now.

PR #461, opened 2018-11-22
Measure text: improve appearance & support RTL

breville:
- Colour goes from pure black to what we use elsewhere on the page for black.
- Text is now equidistant from top and left edges.
- A light rectangle behind the text helps it to be seen against all backgrounds.  (The rectangle has dynamic width based on text width.)
- We render properly for both LTR & RTL.  (Detection is done on the `div` containing the dance party.)

### before

screenshot 2018-11-22 12 22 07

#### LTR

screenshot 2018-11-27 12 08 19

#### RTL

screenshot 2018-11-27 12 08 43

### DIFF
--- a/src/p5.dance.js
+++ b/src/p5.dance.js
@@ -1059,7 +1059,7 @@ module.exports = class DanceParty {
 
       // Background rectangle.
       this.p5_.noStroke();
-      this.p5_.fill("rgba(255,255,255,.8)");
+      this.p5_.fill("rgba(0,0,0,.38)");
       if (this.rtl) {
         this.p5_.rect(399 - 13 - textWidth, 4, textWidth + 10, 28);
       } else {
@@ -1067,7 +1067,7 @@ module.exports = class DanceParty {
       }
 
       // The text.
-      this.p5_.fill("#333");
+      this.p5_.fill("white");
       if (this.rtl) {
         this.p5_.text(text, 399 - 9, 25);
       } else {
--- a/src/p5.dance.js
+++ b/src/p5.dance.js
@@ -55,6 +55,8 @@ module.exports = class DanceParty {
     this.i18n = i18n;
     this.resourceLoader_ = resourceLoader;
 
+    this.rtl = getComputedStyle(document.getElementById(container)).direction === "rtl";
+
     this.world = {
       height: 400,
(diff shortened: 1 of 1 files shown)


breville:
I had trouble with getting a good-looking text border, but I did get a background rectangle working nicely, if we'd like to incorporate it:

screenshot 2018-11-24 14 54 09
screenshot 2018-11-24 14 54 21

breville:
@joshlory Would be great if you could confirm the additional change also looks good.  Thanks!

breville:
This now fixes https://github.com/code-dot-org/dance-party/issues/459 by adding proper RTL support.

src/p5.dance.js
```diff
@@ -1046,14 +1048,31 @@ module.exports = class DanceParty {
...
 
-    this.p5_.fill("black");
-    this.p5_.textStyle(this.p5_.BOLD);
-    this.p5_.textAlign(this.p5_.TOP, this.p5_.LEFT);
-    this.p5_.textSize(20);
-
     this.world.validationCallback(this.world, this, this.sprites_, events);
     if (this.showMeasureLabel && this.getCurrentMeasure() >= 1) {
-      this.p5_.text(`${this.i18n.measure()} ${Math.floor(Math.max(0, this.getCurrentMeasure()))}`, 10, 20);
+      const text = `${this.i18n.measure()} ${Math.floor(Math.max(0, this.getCurrentMeasure()))}`;
```
breville: @Hamms Should localising this text have been done with token substitution so that it works in all locales?

breville:
@Hamms I'm not sure how the text inside the white rectangle should look in RTL languages, particularly the ordering of text, colon, and number.  Thoughts?

src/p5.dance.js
Hamms: arguably yes, but for something as simple as appending a number to a string I believe this should work the same in all locales

breville:
@Hamms Looks like it works pretty well in RTL:

screenshot 2018-11-30 13 35 42

Hamms:
Oh, nice! Yeah, that looks exactly correct! That's cool, I wasn't expecting the browser to know that it's supposed to apply RTL flow to text within canvas

Issue #454, opened 2018-11-21
Background effects should have a palette option


joshlory:
Tracking issue.

simplycloud:
this is done!

PR #463, opened 2018-11-26
Add Mark's initial effect color palettes

joshlory:
Progress on issue https://github.com/code-dot-org/dance-party/issues/454.

I think we may want to update these with more colors, or interpolate colors to get additional variants.  With only 5 colors, we end up with a lot of repeats in e.g. the "disco" background.

From Mark:

dp_colorschemes


simplycloud:
@joshlory we have one additional black / grey / white palette to add.

joshlory:
@islemaster I added commit efa0823cc759d8b4bae48f0d28f39d574a94475f to always re-init the background when changing effects.

PR #458, opened 2018-11-21
new splatters appear on top

Bjvanminnen:
Fix for https://github.com/code-dot-org/dance-party/issues/215

Newer bubbles show up on top of older ones now.

Issue #452, opened 2018-11-21
Falling emoji backgrounds perform poorly on mobile Safari


islemaster:
Initial results from stress testing against the new backgrounds show performance issues in a few of them - particularly ones where we're drawing lots of 'emoji' to the screen:

- raining_tacos
- falling_rainbows
- pizzas
- music_notes
- pineapples
- hearts
- falling_poop

![image](https://user-images.githubusercontent.com/1615761/48822319-7a7c6600-ed11-11e8-954c-be9ea39e5c2a.png)

One of the things they have in common is that they're repeating the same drawing operations many times instead of drawing once to an offscreen canvas and copying to the playspace.  We think this will be a fairly straightforward fix.

breville:
Is there a reason we don't we just load these images from a small PNG spritesheet?

simplycloud:
this is addressed in PR https://github.com/code-dot-org/dance-party/pull/464

simplycloud:
this is now resolved.

PR #464, opened 2018-11-26
new backgrounds / foregrounds and optimizations

simplycloud:
new backgrounds / foregrounds and optimizations. full list of changes:
- new disco ball background (paired with @joshlory)
- new circles background (copied from diamonds background)
- clean up of sparkles background so that sparkles fill more of canvas
- optimization of draw calls for pizza, pineapple, poop, hearts, rainbows, and tacos
- addition of smiley face background
- addition of lasers background
- optimizations of splatter background (ref: https://github.com/code-dot-org/dance-party/pull/458/)
- removal of strobe lights background


simplycloud:
looks like the splatter test failed; which is expected because of a planned change to the baseline. will resolve this in the AM.

src/Effects.js
```diff
@@ -237,34 +329,38 @@ module.exports = class Effects {
...
             rot: randomNumber(0, 359),
-            speed: randomNumber(2, 5),
+            speed: 3,
+            size: 5,
           });
         }
+        this.image = p5.createGraphics(70, 50);
+        this.image.pixelDensity(1);
+        this.image.scale(4,4);
+        drawTaco(this.image.drawingContext);
```
islemaster: :+1: Looking forward to seeing what a difference this makes in performance!

Let's make sure we test this in Safari, since it has known issues with `createGraphics`.  I did investigate a little and I don't see a way to inject our own offscreen canvas context into p5 for drawing - I suspect this will be okay just happening on init with a fairly small canvas, but want to validate before we ship the associated blocks.

src/Effects.js
```diff
@@ -274,75 +370,33 @@ module.exports = class Effects {
...
-        for (let i = 0; i < this.pineapple.length; i++) {
+        for (let i = 0; i < this.pineappleList.length; i++) {
           p5.push();
-          const pineapple = this.pineapple[i];
+          const pineapple = this.pineappleList[i];
           p5.translate(pineapple.x, pineapple.y);
           p5.rotate(pineapple.rot);
-          p5.scale(2, 2);
+          p5.scale(pineapple.life / 20);
           drawPineapple(p5._renderer.drawingContext);
```
islemaster: Pineapples was identified as one of the slow backgrounds in https://github.com/code-dot-org/dance-party/issues/452.  It looks like you chose not to use the draw-once-then-copy strategy here.  Is there a particular reason why?

src/Effects.js
```diff
@@ -479,6 +537,36 @@ module.exports = class Effects {
...
+          y: 1700,
+          z: 400,
+          color: randomColor(255,255,255),
+        };
+        this.laser.push(laser);
+        p5.stroke('white');
+        p5.line(0,200,400,200);
+        this.laser.forEach(laser => {
+          p5.push();
+          p5.translate(this.laser.w, this.laser.x, this.laser.y, this.laser.z);
```
islemaster: What does this do? [`p5.translate`](https://p5js.org/reference/#/p5/translate) is only documented as taking three arguments.

src/Effects.js
```diff
@@ -479,6 +537,36 @@ module.exports = class Effects {
...
+          for (let i = 0; i < 1; i++) {
```
islemaster: Uh... Can we just remove this loop?

src/Effects.js
```diff
@@ -479,6 +537,36 @@ module.exports = class Effects {
...
+      draw: function () {
+        p5.background('black');
+        let laser = {
+          w: 200,
+          x: 200,
+          y: 1700,
+          z: 400,
+          color: randomColor(255,255,255),
+        };
+        this.laser.push(laser);
```
islemaster: :-1: Memory leak. We're pushing a new object into `this.laser` every frame and never removing any of them.  Especially considering we don't reset background states across run/reset this could crash the page across enough runs.

src/Effects.js
```diff
@@ -627,61 +715,54 @@ module.exports = class Effects {
...
           p5.rotate(poop.rot);
-          p5.scale(2, 2);
-          drawFallingPoop(p5._renderer.drawingContext);
-          p5.pop();
-          poop.y += poop.speed;
-          poop.rot++;
-          if (poop.y > 410) {
+          p5.scale(poop.life / 20);
+          p5.drawingContext.globalAlpha - 0.5;
+          drawPoop(p5._renderer.drawingContext);
```
islemaster: Another place we haven't switched to draw-once.  I assume this is because we reported about ~23fps on this effect before and we've cut the number of emoji being drawn by 70%.  This might be 'good enough' but remember that those perf tests are from _just_ running the background, with no dancers onscreen.  If we can drive the cost of the background down further we should.

src/Effects.js
```diff
@@ -691,51 +772,90 @@ module.exports = class Effects {
           const heart = this.heartList[i];
           p5.translate(heart.x, heart.y);
           p5.rotate(heart.rot);
-          p5.scale(2, 2);
-          drawHeart(p5._renderer.drawingContext);
-          heart.y += heart.speed;
-          heart.rot++;
-          if (heart.y > 410) {
+          p5.scale(heart.life / 20);
+          p5.drawingContext.globalAlpha - 0.5;
```
islemaster: I'm pretty sure this line is a no-op.

src/Effects.js
```diff
@@ -691,51 +772,90 @@ module.exports = class Effects {
...
+        if (this.heartList.length < 1) {
+          this.init();
+        }
+        for (let i = 0; i < this.heartList.length; i++) {
+          p5.push();
+          const heart = this.heartList[i];
+          p5.translate(heart.x, heart.y);
+          p5.rotate(heart.rot);
+          p5.scale(heart.life / 20);
+          p5.drawingContext.globalAlpha - 0.5;
```
islemaster: Same no-op as above.

src/Effects.js
```diff
@@ -867,31 +987,69 @@ module.exports = class Effects {
...
-        let strobe = {
-          lineH: 25,
-          lineW: 1,
-          lineX: 0,
-          lightH: 50,
-          lightD: 40,
-        };
-        this.strobe.push(strobe);
-        this.strobe.forEach(function (strobe) {
+        for (let i = 0; i < 1; i ++) {
```
islemaster: Can we just remove this loop too?

src/Effects.js
```diff
@@ -867,31 +987,69 @@ module.exports = class Effects {
...
+        this.pizza.forEach(function (pizza){
           p5.push();
-          for (var x = 50; x <= 400; x += 100) {
-            p5.fill(p5.random(255), p5.random(255), p5.random(255));
-            p5.ellipse(x, strobe.lightH, strobe.lightD, strobe.lightD);
-          }
-          for (var y = 50; y <= 400; y += 100) {
-            p5.rect(y, strobe.lineX, strobe.lineW, strobe.lineH);
+          p5.translate(pizza.x, pizza.y);
+          drawPizza(p5._renderer.drawingContext);
```
islemaster: :-1: Pizzas was one of the worst-performing effects in our tests, and here we are still performing the expensive emoji draw every time.  It's also not clear what the maximum number of pizzas onscreen is since we're pushing a new one every frame and removing them whenever they are out of frame.

src/Effects.js
```diff
@@ -867,31 +987,69 @@ module.exports = class Effects {
...
+          let speedMultiplier = p5.pow(pizza.size, 2) / 8;
+          pizza.x += pizza.velocity.x * speedMultiplier;
+          pizza.y += pizza.velocity.y * speedMultiplier;
```
islemaster: I'm pretty sure `speedMultiplier` is less than 1 for the first several frames, so we're actually slowing the pizza down?

src/Effects.js
```diff
@@ -867,31 +987,69 @@ module.exports = class Effects {
...
+          pizza.size += 0.1;
+          p5.pop();
+        });
+        this.pizza = this.pizza.filter(function (pizza) {
```
islemaster: Smaller issue: This allocates a new array every frame, not the most efficient but probably not a big deal compared to the draw calls.

src/Effects.js
```diff
@@ -867,31 +987,69 @@ module.exports = class Effects {
...
+          if (pizza.x < -5 || pizza.x > 405 || pizza.y < -5 || pizza.y > 405) {
+            return false;
           }
+          return true;
```
islemaster: Code style nit:
```js
return !(pizza.x < -5 || pizza.x > 405 || pizza.y < -5 || pizza.y > 405);
```

src/Effects.js
```diff
@@ -867,31 +987,69 @@ module.exports = class Effects {
...
+        });
+      }
+    };
+
+    this.smile_face = {
+      smile: [],
+      draw: function () {
+        p5.background("#D3D3D3");
+        for (let i = 0; i < 1; i ++) {
```
islemaster: Another no-op loop.

src/Effects.js
```diff
@@ -867,31 +987,69 @@ module.exports = class Effects {
...
+            velocity: p5.createVector(0, 1).rotate(p5.random(0,360)),
+            size: 0.1,
+          };
+          this.smile.push(smile);
+        }
+        p5.noStroke();
+        this.smile.forEach(function (smile){
+          p5.push();
+          p5.translate(smile.x, smile.y);
+          drawSmiley(p5._renderer.drawingContext);
```
islemaster: Similar potential issues to pizza.

src/Effects.js
```diff
@@ -20,6 +20,75 @@ module.exports = class Effects {
...
+          p5.rect(199.5, 0, 2, 10.25);
+          for (let j = 0; j < 180; j += step) {
+            p5.push();
+            p5.fill(p5.noise(i, j, p5.frameCount / 50) * 255 + 100);
+            this.quad(i, j, step, p5.frameCount * 3);
+            p5.pop();
+          }
+        }
+      },
+    };
```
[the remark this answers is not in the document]
joshlory: I did some manual testing of the disco ball and I think we're in a good place, even testing at 6x slowdown.  We can change the `const step = 20` to a larger value to reduce the number of faces (faces scale with the inverse square of the step size).

Maybe we should copy the star graphic to a temp canvas, and blit that whenever we need to draw a star?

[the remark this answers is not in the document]
simplycloud: agreed to push and test performance and then optimize further.

src/Effects.js
simplycloud: @islemaster cleaning these up as part of later merge commit

src/Effects.js
simplycloud: @islemaster cleaning these up as part of later merge commit

src/Effects.js
simplycloud: @islemaster cleaning these up as part of later merge commit. we are going to set the limit to 100 lasers at any given time.

src/Effects.js
simplycloud: great catch; i will clean these up.

.travis.yml
```diff
@@ -1,4 +1,5 @@
 language: node_js
+dist: xenial
```
joshlory: @islemaster do you remember why we made this change?

islemaster: I don't! I did not take good notes in this commit, and it's unclear from the previous test failure why changing distros would fix anything.

If I had to guess, I'd say it's going to be an issue with a native dependency not having a prebuilt package for a particular distro at the time, which might no longer be a problem.

What's default on Travis now?  Do we have a reason not to do this?

PR #474, opened 2018-11-26
Update font for text foreground

simplycloud:
fixes issue where font can vary depending on operating system / browser


simplycloud:
closing as we pulled in the font changes through another merge

PR #478, opened 2018-11-27
Effect cleanup

islemaster:
A pure refactor.  Moves the drawing functions to their own files for easier maintenance, and also cleans up a handful of unneeded/no-op lines in Effects.js.


islemaster:
I'd like to merge this out-of-band so anyone else working on effect changes doesn't have to deal with a terrible merge conflict later.  Thoughts?

epeach:
I think merging this out-of-band is the way to go.

PR #495, opened 2018-11-28
Use palette colors on rainbow and stars

epeach:
Using palette colors on rainbow and star backgrounds.

Changes:
*Each star is initialized with a color and that color is used to fill the star when drawn
*The background of the rainbow is set to the middlest color with 0.25 alpha. Each arc (of 7) is drawn at 1/7 lerp of the palette at 0.5 alpha. The 'arcing' is the color of the main arc plus 20 in each R/G/B and at the same alpha as was used in the original rainbow. This is slightly different than the original design to account for moving from HSLA to RGB values.

Note: I was not able to get the p5.color methods of (setRed, setAlpha, . . .) to be recognized so these values are set by manually accessing the array inside p5.color. When time allows, we should move to leveraging the built in setter method for p5.color.

Examples of updated backgrounds below:

Default rainbow:
screenshot from 2018-11-27 16-24-29

Vintage rainbow:
screenshot from 2018-11-27 16-23-59

Default stars:
screenshot from 2018-11-27 16-24-41


islemaster:
Vintage Rainbow! :heart_eyes:

epeach:
@islemaster Good call on the new baselines. I updated the visual tests to handle specifying palettes and recorded baselines for the rainbow with all the palettes.

Issue #469, opened 2018-11-26
Adding more backgrounds causes "random" in the dropdown to break for old solutions


ryansloan:
"Random" was previously selected, but when we added more backgrounds it changed the key:
<img width="822" alt="screen shot 2018-11-26 at 9 30 48 am" src="https://user-images.githubusercontent.com/1669271/49031149-0bae6b00-f15e-11e8-91d6-41d06fb426ae.png">

Random arg is now:
```
["(Random)",
          "['disco', 'color_cycle', 'rainbow', 'diamonds','snowflakes','splatter','swirl','spiral','starspace'][randomInt(0,8)]"]
```

uponthesun:
Talked to @davidsbailey, and he suggested defining a new block to have the updated value, keeping the old one for backwards compatibility, and updating all of the levels. He'd made a similar change here:

https://github.com/code-dot-org/code-dot-org/pull/26206

ryansloan:
@uponthesun We will definitely be changing this list in the future, though we aren't planning to change the user-facing list again in the HOC timeframe (it will change on our development environments, but we aren't planning to ship more than one more revision of the list publicly for HOC). 

I think if it's possible, the best thing to do is to do it the right way and remove the list and random int selection from the key and give it a `rand` key or something like that. Is that a feasible option for Hour of Code? Or do we need to continue to use the enumerated list as key?

uponthesun:
Talked with @ryansloan - I'll try to implement the "right" fix (adding the "rand" key and adding the list of backgrounds into the dance party code) and have that ready for Thursday.

ryansloan:
Spec for what we're going to do here:

1) Ryan will create a new `setBackgroundEffectWithPalette` block on levelbuilder with an option in the list with display name "(Random)" and key "rand"

2) @uponthesun will update `DanceParty.setBackgroundEffect` to check for `effect=="rand"`, and if so select a random background from this hardcoded list:
```
['circles','color_cycle', 'diamonds', 'disco_ball', 'fireworks', 'swirl', 'kaleidoscope', 'lasers', 'splatter', 'rainbow', 'snowflakes', 'text', 'galaxy', 'sparkles', 'spiral', 'disco', 'stars']
```
I think @islemaster would be a great person to pair with on this, he's done a bunch of work in this repo. Let me know if you have any questions!

ryansloan:
(Sorry, deleted an earlier comment while we iterated on the plan)
Spec for what we’re going to do here:
[...]

PR #499, opened 2018-11-28
New effect commands with 'rand' support

islemaster:
Part of the fix for https://github.com/code-dot-org/dance-party/issues/469

Introduces `setBackgroundEffectWithPalette` and `setForegroundEffectExtended` which are pass-throughs to the existing functions, but exist so that we can create a new block instead of updating the old one for backwards-compatability.

Adds a random effect key 'rand' that, if passed, will cause a random effect to be selected from a preselected list appropriate to background/foreground as appropriate.


src/p5.dance.js
```diff
@@ -1030,6 +1036,16 @@ module.exports = class DanceParty {
...
   }
 
+  /**
+   * Randomly pick one element out of an array.
+   * @param {Array.<T>} collection
+   * @returns {T}
+   * @private
+   */
+  sample_(collection) {
+    return collection[Math.floor(this.p5_.random(0, collection.length))];
```
uponthesun: Slightly simpler to use the randomInt function already defined in this file?

src/api.js
```diff
@@ -6,12 +6,18 @@ module.exports = class DanceAPI {
...
+      setBackgroundEffect: (effect, palette) => {
+        nativeAPI.setBackgroundEffect(effect.toString(), palette);
+      },
+      setBackgroundEffectWithPalette: (effect, palette) => {
+        nativeAPI.setBackgroundEffect(effect.toString(), palette);
       },
       setForegroundEffect: effect => {
         nativeAPI.setForegroundEffect(effect.toString());
       },
+      setForegroundEffectExtended: (effect) => {
```
uponthesun: Do we need this new function in the API if the signature / behavior is the same? Can't both the old and new blocks call the same existing function ?

src/constants.js
```diff
@@ -29,4 +29,39 @@ module.exports = {
     {name: "XOle", mirror: false, shortBurst: true},
     {name: "XSlide", mirror: false, shortBurst: true},
   ],
+  RANDOM_EFFECT_KEY: 'rand',
+  BACKGROUND_EFFECTS: [
```
uponthesun: Is it possible / realistic that this list gets out of sync and contains a background which doesn't exist yet / no longer exists? Where's the authoritative source of implemented backgrounds? If so, is it possible to fail gracefully if we randomly pick one which doesn't exist?

test/unit/apiTest.js
```diff
@@ -0,0 +1,58 @@
+const sinon = require('sinon');
+const test = require('tape');
+const DanceAPI = require('../../src/api');
+
+const FAKE_EFFECT = 'fake-effect-name';
+const FAKE_PALETTE = 'fake-palette-name';
+
+// DEPRECATED
```
uponthesun: Add these comments to the api source file as well?

src/api.js
islemaster: @ryansloan mentioned there was some reason we can't have two different blocks that point to the same function.  I don't know the details, but providing an alias function in the API seemed like a good workaround.

I'm not actually sure what that limitation is - we use the full block name when serializing to XML (`<block type="Dancelab_setBackgroundEffect">`), so there must be some other place we depend on the `func` being unique across all loaded blocks.

src/constants.js
islemaster: That's a great point.  `Effects.js` contains the actual implemented effects.  I'll union this list of 'approved' effects for random selection with the set of implemented effects before we pick one.

src/api.js
ryansloan: The blocks editor enforces uniqueness, I'm not sure how much of that is just the editor vs. some other underlying requirement

PR #501, opened 2018-11-28
Palette disco ball

epeach:
disco_ball

Changes:
*Disco Ball background is a horizontal gradient of the chose palette.
*Using @joshlory 's code to populate (and fade) stars further down the screen

Note: Worked with @simplycloud to adjust the values for the height to populate stars to and when to fade them.

To explore: Coloring the disco ball


ryansloan:
this looks awesome!

src/Effects.js
```diff
@@ -91,24 +91,33 @@ module.exports = class Effects {
...
         }
       },
       draw: function () {
-        p5.background("#9370DB");
+        p5.noFill();
+        // Draw a horizontal gradient of the palette colors to the background
+        for (let i = 0; i <= 425; i++) {
+          let c = lerpColorFromPalette(i / 425);
+          p5.stroke(c);
+          p5.line(0, i, 425, i);
```
joshlory: I just saw this is doing a line per row, instead of a [linear gradient](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/createLinearGradient).  I'm a bit worried about the perf of this change.  This is already one of our more CPU-intensive effects.

PR #518, opened 2018-11-28
Fix to circles effect

simplycloud:
update circle effect to not have a "un-filled" center. adding new circle tests.

circle-v2


joshlory:
I think you'll want to commit the updated `test/visual/fixtures/circles.png` (right now it's just deleted without replacement).

Otherwise LGTM!

PR #525, opened 2018-11-29
Disco ball shading

joshlory:
Small change to add a slight star twinkle, and choose disco ball face colors that look like they could be reflecting the environment.

image


epeach:
Looks like the baseline disco_ball gradient background is lighter (less saturated). Is that intended?

joshlory:
I think it's a side-effect of drawing the gradient with lines.  Once your other change goes in, this should go back to normal (we'll need to update the baseline when both go in).

PR #526, opened 2018-11-29
fix to taco and rainbow canvas size

simplycloud:
fixes to two foregrounds so that emojis properly fit in canvas.

taco-regression

taco-fix

rainbow-before

rainbow-after


simplycloud:
will work with @epeach to get new test baselines generated (or if i can fix canvas on my client, I will update myself). thanks!

simplycloud:
@epeach the updated tests look good! thanks!

PR #533, opened 2018-11-29
Draw pineapple graphic to a temp drawing buffer

joshlory:
Fix for slow pineapple effect performance on iPhone 5s.

image

### DIFF
--- a/src/Effects.js
+++ b/src/Effects.js
@@ -466,6 +466,9 @@ module.exports = class Effects {
             life: 5,
           });
         }
+        this.image = p5.createGraphics(75, 130);
+        this.image.scale(7);
+        drawPineapple(this.image.drawingContext);
       },
       draw: function () {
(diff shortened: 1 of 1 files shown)


src/Effects.js
```diff
@@ -466,15 +466,18 @@ module.exports = class Effects {
             life: 5,
           });
         }
+        this.image = p5.createGraphics(75, 130);
+        this.image.scale(7);
```
simplycloud: we set this scale a lot lower in the other emoji based foregrounds. have we tested this with music to see if the pineapples get to big?

joshlory: Good question!  I went with a higher number here to get more detail in the larger pineapples — the size of the drawn sprites shouldn't change.  The 7 is reduced back down by 1/7 on [line 479](https://github.com/code-dot-org/dance-party/pull/533/files#diff-c91a4859bf33bf9cb2bd5e367b7555a3R479).

PR #544, opened 2018-12-04
expose palettes as a constant, along with background and foreground effects

Hamms:
For use in testing video generation.

We should also consider updating the blocks that consume these to be js-defined rather than levelbuilder-defined, and have them directly reference these constants.


joshlory:
@Hamms is this still active?

Hamms:
Oh, yeah, I still want this for tests in p5-replay. Thanks for finding it!

PR #545, opened 2018-12-04
Moving shapes background effect

joshlory:
image


src/Effects.js
```diff
@@ -668,6 +668,55 @@ module.exports = class Effects {
...
+      init: function () {
+        if (this.buffer) {
+          return;
+        }
+        this.buffer = p5.createGraphics(400, 400);
+        this.buffer.noFill();
+        this.buffer.stroke('#0f0');
+        this.buffer.strokeWeight(2);
+        this.buffer.strokeJoin(p5.BEVEL);
+        this.buffer.background(0);
```
joshlory: As future cleanup, I wonder if we should start sharing this buffer across all the effects?  So that using all the effects won't allocate multiple 400x400 canvas buffers.

islemaster: Maybe?  I'd almost prefer that we had a cleanup step for backgrounds so they could throw away the unused buffer, leading to lower memory footprint when a background that didn't need it was running.  That'd also ensure we don't accidentally share this buffer between, for example, background and foreground effects, and could in theory run an arbitrary number of foreground effects at the same time.

PR #547, opened 2018-12-06
Update Alpha Setting for Color

epeach:
To set the alpha of the palette colors in the rainbow, I previously accessed the array of RGBA values in the p5.Color object. This update removes that 'hacky' method and instead creates a new p5.color object with the same RGB values as the old color and a manually set alpha value.


src/Effects.js
```diff
@@ -58,6 +58,11 @@ module.exports = class Effects {
       return p5.lerpColor(p5.color(prev), p5.color(next), remainder);
     };
 
+    /* Given a p5.Color object, creates a new color object with the given alpha value*/
+    const setAlphaForColor = (color, alpha) => {
+      return p5.color(color._getRed(), color._getGreen(), color._getBlue(), alpha);
```
joshlory: Can we use `p5.red(color)` etc. here?

epeach:
@joshlory How bad was the performance degradation? I can definitely understand that this change would cause that and depending on the impact, it might be worth exploring an alternative workaround.

PR #556, opened 2019-01-19
Add Flower background

ajpal:
screen shot 2019-01-18 at 6 20 21 pm


joshlory:
This looks awesome! @epeach I'm seeing a few other background baselines modified in this PR, do you know why this might be?

ajpal:
Those were generated when I ran the screenshot test. I'm not sure exactly why it made new screenshots for the rainbow background, too, but from looking at the diffs, the images are almost identical.

epeach:
It looks like the background color has changed - not sure yet what would cause that, but I'd say these are legitimate differences.

epeach:
(More update) Looks like I introduced the rainbow change with this PR: https://github.com/code-dot-org/dance-party/pull/547 so I'll handle reverting/updating those changes. If you remove those files from this PR, then it's non-blocking for this change.

PR #559, opened 2019-07-31
added music_wave background

indhu-gunda:
I added a music wave background that pulses with the music. 
Screen Shot 2019-07-31 at 4 56 18 PM


joshlory:
This is so cool!

giphy

src/Effects.js
```diff
@@ -1445,6 +1445,40 @@ module.exports = class Effects {
         }
       }
     };
+
+    this.music_wave = {
+      wave: [],
```
epeach: It looks like we'll only ever have one wave - would it make sense to update this to be a single variable and not an array?
You should be able to do that my replacing `wave: []` with `wave: null` or an empty object and then updating lines 1459 and 1467 to assign and reference `this.wave` as an object.

src/Effects.js
```diff
@@ -1445,6 +1445,40 @@ module.exports = class Effects {
...
+        this.wave.push(wave);
+      },
+
+      draw: function (context) {
+        const centroid = context.centroid;
+        let scale = p5.map(centroid, 5000, 8000, 0, 1);
+        let a = 0;
+        p5.background('black');
+        const wave = this.wave[0];
+        for (let i = 1; i <= wave.colorUnit; i+=5) {
```
epeach: Is there a reason for starting this for-loop at `i=1` instead of `i=0`? In general, I think it's preferable to start at index 0.

src/Effects.js
```diff
@@ -1445,6 +1445,40 @@ module.exports = class Effects {
...
+          p5.stroke(lerpColorFromPalette(i/wave.colorUnit));
+          let yInitial = wave.yHalf - Math.abs(wave.yHalf * scale * (wave.heightFactor/wave.colorHalf) * p5.cos(a));
```
epeach: Style suggestion: If we extract `Math.abs(wave.yHalf * scale * (wave.heightFactor/wave.colorHalf) * p5.cos(a)` into it's own variable, it will be easier to update in the future and, in my opinion, the code becomes more readable.

src/Effects.js
```diff
@@ -1445,6 +1445,40 @@ module.exports = class Effects {
...
+          let yFinal = wave.yHalf + Math.abs(wave.yHalf * scale * (wave.heightFactor/wave.colorHalf) * p5.cos(a));
+          p5.line(i, yInitial, i, yFinal);
+          if ( i > wave.colorHalf) {
```
epeach: From my interpretation, `wave.colorHalf` represents half of `wave.colorUnit`. If we can derive `wave.colorHalf` with `wave.colorUnit/2`, then I think we can get rid of `wave.colorHalf`.

src/Effects.js
```diff
@@ -1445,6 +1445,40 @@ module.exports = class Effects {
...
       }
     };
+
+    this.music_wave = {
+      wave: [],
+      init: function () {
+        let wave = {
+          inc: 360/15,
+          heightFactor: 1,
+          yHalf: 300,
```
epeach: The name of this variable, `yHalf`, makes it seem as though it represents half of the total height of the screen, which I believe would have a value of 200. Can we update this variable name so it doesn't imply that it is half the height?

demo.js
```diff
@@ -43,7 +43,7 @@ const runCode = async function () {
 }
 
 textareaCode.value = textareaCode.value || `var cat = makeNewDanceSprite("CAT", null, {x: 200, y: 200});
-setBackgroundEffectWithPalette("disco_ball", "rand");
```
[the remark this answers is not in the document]
indhu-gunda: Sorry about that! I changed it back to the original.

package-lock.json
```diff
@@ -2812,15 +2815,6 @@
       "integrity": "sha512-xDVs8pBFr6bzq9pXUkLKpGQQnzsF/l6/yX38UnCkTcUcwC0rDl1NGZGildcJVTU+uGBxfsyniK/ZWagPNn1Oqw==",
       "dev": true
     },
-    "canvas": {
```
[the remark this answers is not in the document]
indhu-gunda: I have merged with master and resolved merge conflicts.

src/Effects.js
indhu-gunda: I made wave an object.

src/Effects.js
indhu-gunda: I'm not entirely sure what the problem is when I make `i=0`. When using the p5 editor it was fine, but when testing it locally `i=0` caused the wave to look drastically different and get cut off. That's why I left it as `i=1`.

src/Effects.js
indhu-gunda: I made an `amplitude` variable for it.

src/Effects.js
indhu-gunda: I changed some of the variable names to hopefully increase clarity.

src/Effects.js
indhu-gunda: I realize now that the variable was to mark the y location of the center of the wave. I updated the name to reflect this.

src/Effects.js
```diff
@@ -1445,6 +1445,40 @@ module.exports = class Effects {
...
+        };
+      },
+
+      draw: function (context) {
+        const centroid = context.centroid;
+        let scale = p5.map(centroid, 5000, 8000, 0, 250);
+        let angle = 0;
+        p5.background('black');
+        const wave = this.wave;
+        for (let i = 1; i <= wave.xTotal; i+=5) {
```
epeach: Nit-pick (i.e. not worth blocking this from shipping, but worth thinking about) - It looks like the way this loop works is that it is 'looping over' the pixel width of the p5 draw space, drawing a new line every five pixels. It might be clearer to 'loop over' the number of lines instead (something like `for (let i = 0; i < 80; i++) {`). Either way should make the same drawing, so it just depends on what makes the most sense to you.

indhu-gunda: Since the value of i is used when drawing the line, I used a variable for the lineWidth. I think this makes the code more flexible to changes in the size of the drawing area.

src/Effects.js
```diff
@@ -1445,6 +1445,40 @@ module.exports = class Effects {
         }
       }
     };
+
+    this.music_wave = {
+      wave: {},
+      init: function () {
```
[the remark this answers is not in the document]
indhu-gunda: I made this change.

src/Effects.js
```diff
@@ -1445,6 +1445,40 @@ module.exports = class Effects {
...
+        this.wave = {
+          inc: 360/15,
+          heightFactor: 1,
+          heightDivider: 200,
+          yLoc: 300,
+          xTotal: 400,
```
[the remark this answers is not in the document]
indhu-gunda: Thank you for the recommendation. That makes a lot more sense.

src/Effects.js
```diff
@@ -1445,6 +1445,35 @@ module.exports = class Effects {
         }
       }
     };
+
+    this.music_wave = {
+      inc: 360/15,
+      heightFactor: 1,
+      heightDivider: 200,
+      yLoc: 300,
+      xTotal: p5.width,
```
epeach: Nit: Since `xTotal` now just holds the value of `p5.width`, we don't have to save the variable `xTotal` and just replace references to it with p5.width.

indhu-gunda: I believe it is all fixed now. The PR only shows Effects.js as modified. I also made the above change.

src/Effects.js
```diff
@@ -1445,6 +1445,34 @@ module.exports = class Effects {
...
       }
     };
+
+    this.music_wave = {
+      inc: 360/15,
+      heightFactor: 1,
+      heightDivider: 200,
+      yLoc: 300,
+      lineWidth: p5.width/80,
+      angle: 0,
```
epeach: One last clean-up! It doesn't look like we use this angle variable but instead use the angle variable on line 1459. If that's correct, let's get rid of this variable.

indhu-gunda: Sorry about that. Seems like I left it out when changed the type of variable.

ajpal:
Awesome! I'm so excited for this background!!

PR #561, opened 2019-08-02
Add Starz background

azhou4:
starz

### DIFF
--- a/src/Effects.js
+++ b/src/Effects.js
@@ -10,6 +10,7 @@ const drawSparkle = require('./shapes/sparkle');
 const drawSpiral = require('./shapes/spiral');
 const drawSwirl = require('./shapes/swirl');
 const drawTaco = require('./shapes/taco');
+const drawStar = require('./shapes/star');
 
 module.exports = class Effects {
   constructor(p5, alpha, blend, currentPalette = 'default') {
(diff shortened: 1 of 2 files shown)


src/Effects.js
```diff
@@ -1445,6 +1474,48 @@ module.exports = class Effects {
         }
       }
     };
+    
+    this.starz = {
```
ajpal: It looks like you've ended up with two versions of your background. I think you can delete this one.

src/Effects.js
```diff
@@ -57,6 +58,34 @@ module.exports = class Effects {
       }
     };
 
+    this.starz = {
+      starSizes: [10, 9, 8, 7, 6, 5, 4, 3, 2, 1],
```
ajpal: Instead of having the values from 1-10 stored in an array, can you restructure your loop to use a counter? Something like: `for (let size = 10; i > 0; i--) {...}`

src/Effects.js
```diff
@@ -57,6 +58,34 @@ module.exports = class Effects {
...
+          let color = lerpColorFromPalette(ceil(size) % 4);
+          push();
+          scale(size);
+          fill(color);
+          drawStar(0, 0, 30, 70, 5);
+          pop();
+        }
+      },
+
+      setup: function() {
```
ajpal: I don't think you need this. The canvas size and frame rate are the same for all background effects.

PR #564, opened 2019-09-06
Create emoji background

yugreta:
Adding a new foreground with various falling emojis
Screen Shot 2019-10-24 at 8 54 03 PM


src/Effects.js
```diff
@@ -1445,6 +1449,67 @@ module.exports = class Effects {
...
+        if (p5.frameCount % 10 === 0) {
+          this.emojiList.push(new this.emoji(this.emojiTypes, this.emojiList));
+        }
+        for (let aEmoji of this.emojiList) {
+          aEmoji.update();
+          aEmoji.display();
+        }
+      },
+
+      emoji : function (emojiTypes, emojiList) {
```
epeach: To simplify the code and make it more readable, I don't think we need the 'emoji' function. If we store the `x`, `y`, `size`, and `image` in an object, similar to how we initialize smiles here: https://github.com/code-dot-org/dance-party/blob/083579e8dec56373028c8b935e089fd5b40c32d5/src/Effects.js#L1340, and combine the update and display code into the `draw` function, similar to how we draw smiles here: https://github.com/code-dot-org/dance-party/blob/083579e8dec56373028c8b935e089fd5b40c32d5/src/Effects.js#L1355, we can get rid of the emoji function. This will make this effect more similar to the other foreground effects and improve the readability.

src/Effects.js
```diff
@@ -1445,6 +1449,67 @@ module.exports = class Effects {
...
+        this.x = randomNumber(0, 350);
+        this.y = -50;
+        this.size = randomNumber(25, 50);
+        let emojiImage = emojiTypes[randomNumber(0, 4)];
+
+        this.update = function () {
+          this.y += p5.pow(this.size, 0.25);
```
epeach: Can we add a short comment here explaining that the 0.25 is the fall-rate? That will make it easy in case we want to update this in the future.

src/Effects.js
```diff
@@ -1445,6 +1449,67 @@ module.exports = class Effects {
...
+          if (this.y > p5.height * 1.2) {
```
epeach: Similar to explaining the 0.25, can we add a short comment explaining the 1.2 value?

src/Effects.js
```diff
@@ -1445,6 +1449,67 @@ module.exports = class Effects {
...
+        this.emojiTypes.push(this.imageTickled);
+
+        this.imageWink = p5.createGraphics(100, 100);
+        this.imageWink.scale(3);
+        drawWink(this.imageWink.drawingContext);
+        this.emojiTypes.push(this.imageWink);
+      },
+
+      draw : function () {
+        if (p5.frameCount % 10 === 0) {
```
epeach: I think it would be useful to add a comment here explaining that a new emoji is generated every 10 frames. This will improve the readability of the code and make it easy to update in case we want to make changes in the future.

src/Effects.js
yugreta: Good idea! I'll do that.

epeach:
Can you add a small description to this PR? Something like, "Adding a new foreground with falling emojis . . ." and a screenshot of the foreground running, would be perfect - thank you! You can see an example of a PR description I wrote in this PR: https://github.com/code-dot-org/dance-party/pull/567

PR #567, opened 2019-09-25
New Emoji Shapes

epeach:
Adding four new svg emojis: heart eyes, star eyes, tickled face, and winking with a tongue out!
These are planned to be used in the emoji raining background: https://github.com/code-dot-org/dance-party/pull/564

These images were taken by substituting each emoji for the hearts in the foreground:
lovestruck
starstruck
tickled
wink


ajpal:
Can you attach screenshots of what these look like? Also out of curiosity- how did you make these? (I'm assuming not by hand?)

src/shapes/lovestruck.js
```diff
@@ -0,0 +1,72 @@
...
+  ctx.moveTo(0,0);
+  ctx.lineTo(17,0);
+  ctx.lineTo(17,16);
+  ctx.lineTo(0,16);
+  ctx.closePath();
+  ctx.clip();
+  ctx.translate(0,0);
+  ctx.translate(0,0);
+  ctx.scale(1,1);
+  ctx.translate(0,0);
```
joshlory: You can clear out the default clipping and translation steps here if you want to optimize a bit!

Also the save/restore steps aren't always necessary, if the `fillStyle` is being specifically set to a new value. Just make sure they're balanced (equal number of save and restore calls).

epeach:
Screenshots incoming . . .

Good call on documenting how these were made.
Mark created svg files of the new emojis. I then open the svg files in a text editor to get the XML for the files. I copied the XML into this svg to this Svg to Canvas converter: http://www.professorcloud.com/svg-to-canvas/
From there I added the shapes to the codebase

src/shapes/lovestruck.js
epeach: Which ones are the default clipping and translations?

epeach:
@ajpal and @joshlory PTAL - thanks!

PR #568, opened 2019-10-02
Stars Foreground Effect

ajpal:
Credit to Andrea Arias for idea and initial implementation.

exploding_stars


src/Effects.js
```diff
@@ -1254,6 +1255,37 @@ module.exports = class Effects {
       }
     };
 
+    this.exploding_stars = {
+      stars: [],
+      initStars: function () {
```
epeach: If we use `init` as the function name, I don't think we'll then need to call the `this.initStars()` on line 1277.

src/Effects.js
```diff
@@ -1254,6 +1255,37 @@ module.exports = class Effects {
...
+            x: 200,
+            y: 200,
+            dx: velocity * p5.cos(theta),
+            dy: velocity * p5.sin(theta)
+          });
+        }
+      },
+      draw: function () {
+        p5.angleMode(p5.RADIANS);
+        p5.noStroke();
```
epeach: What does this `noStroke` do?

src/Effects.js
```diff
@@ -1254,6 +1255,37 @@ module.exports = class Effects {
...
     };
 
+    this.exploding_stars = {
+      stars: [],
+      initStars: function () {
+        for (let i = 0; i < 100; i++) {
+          let theta = p5.random(0, p5.TWO_PI);
+          let velocity = p5.random(4,12);
+          this.stars.push({
+            color: randomColor(255, 255, 100),
```
epeach: Can we use the color palette to determine the star color? I think we can use 'randomColorFromPalette'.

src/shapes/star.js
```diff
@@ -0,0 +1,15 @@
+// Star helper function from https://p5js.org/examples/form-star.html
+module.exports = function drawStar(p5, x, y, radius1, radius2, numPoints) {
+  let angle = p5.TWO_PI / numPoints;
```
epeach: Do any of the other shapes use p5? Thematically, it makes sense to put this here, but I believe it breaks an established pattern.

src/Effects.js
ajpal: The foreground block doesn't have a palette, and it seemed like none of the other foreground effects use the palette. I could change it to match the selected palette if there is one, but if there's no palette selected fall-back to random? That seems a little weird though because then the foreground block depends on the value from the background block? What do you think?

src/shapes/star.js
ajpal: I guess it could also just be extracted as a helper function within Effects.js, like `lerpColorFromPalette()`. Do you think that would be better?

src/Effects.js
epeach: You're totally right - I just forgot this was a foreground. Ignore me!

src/shapes/star.js
epeach: I like your idea to group them thematically.

src/Effects.js
epeach: Bump

ajpal: Weird, I thought I responded to this, but must not have...
`noStroke()` prevents an outline, like 
image

PR #569, opened 2019-10-03
Squiggles Background Effect

ajpal:
Credit to Gaby Pulido for idea and initial implementation
squiggles


src/Effects.js
```diff
@@ -1445,6 +1445,40 @@ module.exports = class Effects {
...
+      init: function () {
+        this.points = [];
+        p5.noStroke();
+        p5.angleMode(p5.DEGREES);
+        let numPoints = p5.height / this.dotSpacing;
+        for (var i = 0; i < numPoints; i++) {
+          this.points.push({
+            x: 0,
+            y: i * this.dotSpacing,
+            theta: 360 / this.period * i * this.dotSpacing,
```
epeach: What is the theta?

src/Effects.js
```diff
@@ -1445,6 +1445,40 @@ module.exports = class Effects {
...
+      dotRadius: 7,
+      numSquiggles: 8,
+      init: function () {
+        this.points = [];
+        p5.noStroke();
+        p5.angleMode(p5.DEGREES);
+        let numPoints = p5.height / this.dotSpacing;
+        for (var i = 0; i < numPoints; i++) {
+          this.points.push({
+            x: 0,
```
epeach: Do all points start at the bottom of the page?

ajpal: `point.x` is set each tick of the draw loop, so setting to 0 here just to initialize it to something.

src/Effects.js
ajpal: Theta is the current angle of the sine curve for each dot

PR #572, opened 2019-10-04
Look up sprites by name property, not variable reference

ajpal:
Depends on https://github.com/code-dot-org/code-dot-org/pull/31120
Previously, sprites were variables whose value was the id reference provided from native. This changes the generated code and sprite lab library so that sprites are instead identified and referenced by name property. Right now, this is intended to be basically a no-op functionality change. Mostly, that just means that sprite names must be unique and there is no way to change a sprite's name after its creation. I could see either or both of those changing at some point, but for now maintaining existing behavior seemed reasonable.

Similar to https://github.com/code-dot-org/code-dot-org/pull/30979

### DIFF
--- a/src/api.js
+++ b/src/api.js
@@ -1,7 +1,7 @@
 module.exports = class DanceAPI {
   constructor(nativeAPI) {
     const sprites = [];
-    
+
     const lookupSprite = spriteName =>
       sprites.find(sprite => sprite.name === spriteName);
 
--- a/src/api.js
+++ b/src/api.js
@@ -1,6 +1,17 @@
 module.exports = class DanceAPI {
   constructor(nativeAPI) {
     const sprites = [];
+    
+    const lookupSprite = spriteName =>
+      sprites.find(sprite => sprite.name === spriteName);
+
+    const enforceUniqueName = name => {
+      sprites.forEach(sprite => {
+        if (sprite.name === name) {
+          sprite.name = undefined;
+        }
+      });
+    };
 
     return {
       setBackground: color => {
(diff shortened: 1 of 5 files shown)


src/api.js
```diff
@@ -1,6 +1,17 @@
 module.exports = class DanceAPI {
   constructor(nativeAPI) {
     const sprites = [];
+    
+    const lookupSprite = spriteName =>
+      sprites.find(sprite => sprite.name === spriteName);
+
+    const enforceUniqueName = name => {
+      sprites.forEach(sprite => {
+        if (sprite.name === name) {
```
epeach: Can you confirm my understanding - if I create a sprite 'dog1' and create a second sprite 'dog1', this just 'removes' a way to reference the first sprite (so it gets basically overridden)?

src/p5.dance.js
```diff
@@ -352,6 +352,10 @@ module.exports = class DanceParty {
     }
 
     var sprite = this.p5_.createSprite(location.x, location.y);
+    
+    if (name) {
+      sprite.name = name;
```
epeach: Not sure if it's important, but at this point, a sprite isn't valid without a name since we could never reference it? Is there a way to error handle that case?

src/api.js
ajpal: Yep, that's right. This is primarily to match the current behavior, though I could see at some point wanting to allow multiple sprites having the same name.

src/p5.dance.js
ajpal: There is a use case for anonymous dance sprites, since you can still interact with them by costume group. In that case, name would be `null` (see https://github.com/code-dot-org/dance-party/blob/91c92500b91c239ce2d6a90fcc1a555152156bb0/src/api.js#L41)

Issue #570, opened 2019-10-04
Implement `changePropByEach` for changing a property on a group of sprites


joshlory:
The delta should be applied to each sprite individually. So if starting size is [20, 20, 100], a change by 10 would result in [30, 30, 110].

joshlory:
+cc @mikeharv

mikeharv:
[block definition](https://levelbuilder-studio.code.org/pools/Dancelab/blocks/Dancelab_changePropEachBy/edit)

note name: "changePropEachBy"
Hope that isn't confusing!

PR #573, opened 2019-10-18
Add ChangePropEachBy to api and test

epeach:
Per this spec: https://docs.google.com/document/d/1MJzshSBGHx4azMgyJSfmuYll77T8HHodQFMsY_4Pp3w/edit#heading=h.nxq7k4gyikj
This issue: https://github.com/code-dot-org/dance-party/issues/570

With test

### DIFF
--- a/src/p5.dance.js
+++ b/src/p5.dance.js
@@ -871,6 +871,13 @@ module.exports = class DanceParty {
     this.setPropEach(group, "dance_speed", val);
   }
 
+  changePropEachBy(group, property, val) {
+    group = this.getGroupByName_(group);
+    group.forEach(function (sprite){
+      this.changePropBy(sprite, property, val);
+    }, this);
+  }
+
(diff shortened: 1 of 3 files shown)


src/p5.dance.js
```diff
@@ -875,6 +875,13 @@ module.exports = class DanceParty {
     this.setPropEach(group, "dance_speed", val);
   }
 
+  changePropEachBy(group, property, val) {
```
ajpal: Instead of adding a new function to handle the group case, can we just make the `changePropBy` function able to handle either individual sprites or groups? And same for all the other functions that currently have two versions.

epeach: Talked offline. Going to track and follow-up in separate PR: https://codedotorg.atlassian.net/browse/STAR-770

epeach:
@ajpal PTAL - thanks!

Issue #183, opened 2018-11-07
Request: `randomize (all ducks) size`


mikeharv:
This is not a bug so mark it low priority. Beginning in level 11, students get a block to randomize a sprite's property. This is great for `when up pressed` or `every 2 measures`. It feels like it's missing from the set of group commands. I also wonder if we want to be able to `set all ducks tint to blue`.

boatmarina:
good idea.  Pushing this to Post-HOC milestone as a great enhancement idea

joshlory:
One possible conflict: Hadi wants the `"all" do "(random)" dance` to make all of the actors do the _same_ random dance.  This might be at odds conceptually with setting each dancer to a _different_ random location.

mikeharv:
@joshlory Could the `randomize "all" (property)` block work the same as Hadi's suggestion (set all to the same random size/rotation/tint) but we just leave off location? Users can already randomize all locations using the layout blocks.

mikeharv:
[block definition](https://levelbuilder-studio.code.org/pools/Dancelab/blocks/Dancelab_setPropRandomEach/edit)

PR #575, opened 2019-10-18
Add setPropRandomEach to Dance API

epeach:
Per this spec: https://docs.google.com/document/d/1MJzshSBGHx4azMgyJSfmuYll77T8HHodQFMsY_4Pp3w/edit#heading=h.nxq7k4gyikj
This issue: https://github.com/code-dot-org/dance-party/issues/183
With test

### DIFF
--- a/test/unit/groupTest.js
+++ b/test/unit/groupTest.js
@@ -292,3 +292,33 @@ test('LayoutSprites resets rotation', async t => {
 
   nativeAPI.reset();
 });
+
+test('randomizing property of a group sets to individually random properties', async t => {
+  const nativeAPI = await helpers.createDanceAPI();
+  nativeAPI.play({
+    bpm: 120,
+  });
+  nativeAPI.setAnimationSpriteSheet("CAT", 0, {}, () => {});
+  nativeAPI.makeNewDanceSpriteGroup(5, 'CAT', 'circle');
+
+  nativeAPI.layoutSprites('CAT', 'grid');
+  let cats = nativeAPI.getGroupByName_('CAT');
+  for (let i = 0; i < cats.length; i++) {
+    nativeAPI.setProp(cats[i].name, 'scale', 20);
+  }
+
+  nativeAPI.setPropRandomEach(cats, 'scale');
+
+  let catScales = [];
+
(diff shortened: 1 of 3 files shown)


test/unit/groupTest.js
```diff
@@ -292,3 +292,33 @@ test('LayoutSprites resets rotation', async t => {
...
+
+test('randomizing property of a group sets to individually random properties', async t => {
+  const nativeAPI = await helpers.createDanceAPI();
+  nativeAPI.play({
+    bpm: 120,
+  });
+  nativeAPI.setAnimationSpriteSheet("CAT", 0, {}, () => {});
+  nativeAPI.makeNewDanceSpriteGroup(5, 'CAT', 'circle');
+
+  nativeAPI.layoutSprites('CAT', 'grid');
```
ajpal: Is this necessary?

test/unit/groupTest.js
```diff
@@ -292,3 +292,33 @@ test('LayoutSprites resets rotation', async t => {
 
   nativeAPI.reset();
 });
+
+test('randomizing property of a group sets to individually random properties', async t => {
```
epeach: @ajpal 's calculations on the flakiness of this test: `setPropRandom('scale') chooses a random value between 0 and 100. So the likelihood of choosing the same value for all 5 sprites is (1/100)^4 = expected 1 failure every 100 million runs.`

test/unit/groupTest.js
```diff
@@ -301,22 +325,22 @@ test('randomizing property of a group sets to individually random properties', a
...
   for (let i = 0; i < cats.length; i++) {
     nativeAPI.setProp(cats[i].name, 'scale', 20);
   }
 
   nativeAPI.setPropRandomEach(cats, 'scale');
 
-  let catScales = [];
+  // Create a list of cat scales and fail if at least one is not unique
+  // This test will fail when all five scales are randomly changed to
+  // the same value and should be re-run in those flakey cases.
```
ajpal: Can you add to this comment a note that we expect failures to be extremely rare? That way if for some reason we start seeing this test be more flaky than expected we can investigate.

PR #578, opened 2019-10-23
Adding two new background effects: ripples and ripples_random

daynew:
# Description
A new background effect for dance-party. Makes ripples (a growing circle) on every "Peak" in the song. The ripples either start from the center or a random location depending on if "ripples" or "ripples_random" are selected.

# Testing
* `npm run test`
* `npm run dev` 

# Open Questions
* Is it okay that the effect uses timestamps to measure how much time has passed?
* Is this effect interesting enough to include? Is it too noisy?

# Preview
ripples
ripples_random


src/constants.js
```diff
@@ -41,6 +41,8 @@ module.exports = {
     'lasers',
     'splatter',
     'rainbow',
+    'ripples',
```
epeach: Let's leave the ripples out of this constant until we are ready to release it to the public (and we've tested it for seizure concerns).

src/Effects.js
```diff
@@ -277,6 +277,90 @@ module.exports = class Effects {
...
+          this.ripples = this.ripples.filter(function (ripple){
+            return ripple.width < maxRippleWidth;
+          });
+          p5.pop();
+          // keep track of the draw times so we can calculate how much time has passed.
+          this.lastDrawTime = currentTime;
+        },
+
+        // creates a object representing the size and position of a ripple given an initial width
+        createRandomRipple: function (width) {
```
epeach: It looks like this can be used to create a random or non-random ripple. Can we rename it to reflect that?

src/Effects.js
```diff
@@ -277,6 +277,90 @@ module.exports = class Effects {
...
+          // calculate how much the ripples have grown and draw them.
+          let rippleWidthGrowth = this.getRippleGrowth(currentTime, bpm);
+          for (let i=0; i < this.ripples.length; i++) {
+            let ripple = this.ripples[i];
+            ripple.width += rippleWidthGrowth;
+            p5.fill(ripple.color);
+            p5.ellipse(ripple.x, ripple.y, ripple.width);
+          }
+          // remove ripples which are too big
+          this.ripples = this.ripples.filter(function (ripple){
```
epeach: For the ripples that start at the center - we can use a much smaller (like 575) maxWidth (based on the p5.width we use of 400 - we could even calculate a maxWidth based on the p5.width to make this extendable in the future). With the smaller maxWidth we can optimize and not have to draw as many ripples that have extended beyond the screen.

epeach:
Thinking more on this - I wonder if we can use p5.frameCount to replace the timestamp processs?

jmkulwik:
We should make sure we check this for epilepsy risk.

daynew:
> We should make sure we check this for epilepsy risk.

Erin will hopefully have a chance to check them this weekend.

epeach:
Yup! The tool we use requires Windows, so I'll be checking for that this weekend.

PR #580, opened 2019-11-04
Add new dance character spritesheets

epeach:
Followed instructions here: https://github.com/code-dot-org/dance-spritesheets
Added the new character output to the characters.json in alphabetic location.


islemaster:
What did your testing process look like?

PR #582, opened 2019-11-06
Remove deprecated sprite sheet creator

epeach:
Because we have https://github.com/code-dot-org/dance-spritesheets I would like to remove this deprecated file so people don't use it.

Screenshot from 2019-11-06 11-48-42

Part of an effort to clean and simplify the spritesheet creation protocol.


islemaster:
Should we add documentation in this repo that references the preferred sprite-sheet creator?

epeach:
Good call! I'll add something to the README

PR #583, opened 2019-11-06
Move local copies of assets

epeach:
We were only keeping these assets around to test with so I moved them to the test folder. I updated the reference in runLevel.
I also updated the test spritesheets to include the new character.

cleanupcat
Part of a dance party clean-up to simplify how we store, reference, and create spritesheets


joshlory:
One downside: this will make our unit tests depend on the S3 assets. Will require an internet connection to run the integration tests, and changes outside this repo can make tests start failing here.

Should we instead move these files to the `test/` dir, so it's clear they are test collateral and not used in production?

islemaster:
:+1: I think Josh has correctly identified the reason we left these in the repo last year.  +1 to moving them, perhaps to `test/fixtures` or `test/assets`.  Then, it'd be interesting to check whether we could reasonably test with a lot _less_ spritesheets, like only one character or a limited set of moves.

epeach:
Good call on leaving them here so we don't rely on S3 access. I'll make this change and update the documentation in dance-spritesheets repo to indicate that new spritesheets should be added here and in s3.

epeach:
@joshlory / @islemaster PTAL - thanks!

PR #584, opened 2019-11-06
Add new character to the dance party sprite characters

epeach:
cc: @joshlory 
Add the new dance party character to the list of constants so that it becomes available.

Next steps:
Bump the dance party version.
Update references from 'https://s3.console.aws.amazon.com/s3/buckets/cdo-curriculum/images/sprites/dance_20181127/' to 'https://s3.console.aws.amazon.com/s3/buckets/cdo-curriculum/images/sprites/dance_20191106/' so that it includes the new character in the spritesheet.

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -1,7 +1,7 @@
 module.exports = {
   SIZE: 300,
   FRAMES: 24,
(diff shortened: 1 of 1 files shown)


joshlory:
Where does this const get used? Wanna make sure it doesn't start showing up in existing blocks before we're ready.

epeach:
@joshlory This constant is mostly used to preload animations (in p5.dance.js). Until the sloth is specified in the block, like ('Dancelab_setPropRandomEach') it should not show up.

joshlory:
Great, sounds good!

PR #587, opened 2019-11-08
Add growing_stars background

ajpal:
Adds growing_stars background effect. Credit to @azhou4 for idea and initial implementation

Nov-08-2019 11-35-25


epeach:
Add a description, please!

src/Effects.js
```diff
@@ -1541,6 +1541,40 @@ module.exports = class Effects {
...
 
+    this.growing_stars = {
+      stars: [],
+      colorIndex: 0,
+      starSpacing: 30,
+      numStars: 9,
+
+      init: function () {
+        p5.angleMode(p5.RADIANS);
+        this.colorIndex = 0;
```
epeach: Do we need to set these to 0 (and []) in init if they are initialized as such at the top? Or does this have to do with resetting?

src/Effects.js
```diff
@@ -1541,6 +1541,40 @@ module.exports = class Effects {
...
+        this.stars = [];
+        for (var i = 0; i < this.numStars; i++) {
```
epeach: draw loop was something like: 
''for(var i = 0; i < this.numStars; i++){
 p5.fill(colorFromPalette(colorIndex + i));
...
}" but I'm not sure if that buys us any additional efficiency. Not ship-blocking, but something the consider.

ajpal: Sorry not sure I follow?

src/Effects.js
ajpal: Yeah we need to reset these in `init` so that if you run/reset/run it works as expected

src/Effects.js
epeach: Gah - sorry! I'm not sure where the first part of that comment went, but I edited it to include the first part of the sentence. Let me know if it still doesn't make sense and I'll elaborate!

ajpal: Ah ok makes sense. I don't think that works in this case because the index in the array changes as the stars grow (when a star gets too big it's pulled off the front of the array and a new star is added to the back of the array), so it ends up looking kind of choppy:
Nov-08-2019 11-42-08

epeach: Got it - thanks for investigating!

PR #590, opened 2019-11-09
Add alternateMoves library function

ajpal:
Will be used for a new block:
image

Alternates between two moves at a given cadence. Implementing by adding `alternatingMovesInfo` to the sprite object, and checking in the draw loop whether the move should be switched. `alternatingMovesInfo` is reset in `changeMoveLR` so that the sprite stops alternating if another block changes the move.

### DIFF
--- a/src/p5.dance.js
+++ b/src/p5.dance.js
@@ -568,6 +568,15 @@ module.exports = class DanceParty {
     return this.sprites_by_type_[group];
   }
 
+  toggleBetween(group, n, move1, move2) {
+    let currentMeasure = Math.floor(this.getCurrentMeasure());
+    if (Math.ceil(currentMeasure / n) % 2 == 0) {
+      this.changeMoveEachLR(group, move2, -1)
+    } else {
+      this.changeMoveEachLR(group, move1, 1)
+    }
+  }
+
   changeMoveEachLR(group, move, dir) {
     group = this.getGroupByName_(group);
     if ((move === "rand") && (group.length>0)) {
--- a/src/p5.dance.js
+++ b/src/p5.dance.js
@@ -571,9 +571,9 @@ module.exports = class DanceParty {
   toggleBetween(group, n, move1, move2) {
     let currentMeasure = Math.floor(this.getCurrentMeasure());
     if (Math.ceil(currentMeasure / n) % 2 == 0) {
-      this.changeMoveEachLR(group, move2, -1)
+      this.changeMoveEachLR(group, move2, -1);
     } else {
-      this.changeMoveEachLR(group, move1, 1)
+      this.changeMoveEachLR(group, move1, 1);
     }
   }
 
--- a/src/p5.dance.js
+++ b/src/p5.dance.js
@@ -456,8 +456,13 @@ module.exports = class DanceParty {
   // Dance Moves
 
   alternateMoves(group, n, move1, move2) {
+    // Maximum of 10 events per second
+    // Our fastest song is 169bpm
+    // This allows events on eighth-notes in that song.
+    n = Math.max(0.1, n);
+
     group = this.getGroupByName_(group);
-    let currentMeasure = Math.floor(this.getCurrentMeasure());
+    let currentMeasure = this.getCurrentMeasure();
     if (currentMeasure === 0) {
       currentMeasure = 1;
     }
@@ -1157,12 +1162,12 @@ module.exports = class DanceParty {
       title,
     };
 
-    let currentMeasure = Math.floor(this.getCurrentMeasure());
+    let currentMeasure = this.getCurrentMeasure();
     this.sprites_.forEach(sprite => {
       if (sprite.alternatingMoveInfo) {
         let alternatingMoveInfo = sprite.alternatingMoveInfo;
-        let val = Math.floor((currentMeasure - sprite.alternatingMoveInfo.start) / sprite.alternatingMoveInfo.cadence);
-        if (val % 2 == 0) {
+        let quotient = Math.floor((currentMeasure - sprite.alternatingMoveInfo.start) / sprite.alternatingMoveInfo.cadence);
+        if (quotient % 2 == 0) {
           if (sprite.alternatingMoveInfo.current != 1) {
             this.changeMoveLR(sprite, sprite.alternatingMoveInfo.move1, -1);
             sprite.alternatingMoveInfo = alternatingMoveInfo;
--- a/src/p5.dance.js
+++ b/src/p5.dance.js
@@ -455,6 +455,24 @@ module.exports = class DanceParty {
 
   // Dance Moves
 
+  alternateMoves(group, n, move1, move2) {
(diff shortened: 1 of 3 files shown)


ajpal:
Turns out this doesn't work:
image
Expected: start dabbing at measure 3, start clapping at measure 5, dabbing at 7, etc
Observed: start dabbing at measure 3 forever

mikeharv:
I could be wrong, but I'm not sure that the new block is semantically equivalent to the screenshot above. 
At the risk of quickly getting out of my depth, I'd expect it work similar to a single `do forever` block. 

If a student wrote the following:
```cats do floss forever
cats alternate dab and drop
cats do clap high forever```
I think we'd expect to only see the cats clap. Each `alternate` or `do forever` block takes the place of the last one and runs repeatedly like a sprite lab behavior.

The new block is not inherently connected to a an event like `repeat every n measures` but it also could placed in an event like that. I can't think of any actual useful cases off the top of my head, but if a student wrote:

```every 3 measures
cats alternate every 2 measures floss and star```

...I'd personally expect to see the cats floss for 2 measures then star for 1 measure, then repeat.

File: src/p5.dance.js [hunk 1]
```diff
@@ -1138,6 +1157,27 @@ module.exports = class DanceParty {
       title,
     };
 
+    let currentMeasure = Math.floor(this.getCurrentMeasure());
+    this.sprites_.forEach(sprite => {
+      if (sprite.alternatingMoveInfo) {
+        let alternatingMoveInfo = sprite.alternatingMoveInfo;
+        let val = Math.floor((currentMeasure - sprite.alternatingMoveInfo.start) / sprite.alternatingMoveInfo.cadence);
+        if (val % 2 == 0) {
```
jmkulwik: Would it make sense to change this to `if (val % 2 == 0 && sprite.alternatingMoveInfo.current > 2)`  ?

jmkulwik: oh wait. Nevermind

src/p5.dance.js
```diff
@@ -455,6 +455,24 @@ module.exports = class DanceParty {
...
+      currentMeasure = 1;
+    }
+    group.forEach(sprite => {
+      this.changeMoveLR(sprite, move1, -1);
+      sprite.alternatingMoveInfo = {
+        start: currentMeasure,
+        cadence: n,
+        move1: move1,
+        move2: move2,
+        current: 1
```
epeach: Is this necessary or could this be calculated from the start, cadence, and current measure?

src/p5.dance.js
```diff
@@ -455,6 +455,24 @@ module.exports = class DanceParty {
...
+    group = this.getGroupByName_(group);
+    let currentMeasure = Math.floor(this.getCurrentMeasure());
+    if (currentMeasure === 0) {
+      currentMeasure = 1;
+    }
+    group.forEach(sprite => {
+      this.changeMoveLR(sprite, move1, -1);
+      sprite.alternatingMoveInfo = {
+        start: currentMeasure,
+        cadence: n,
```
epeach: Nit - n isn't a very descriptive variable name.

src/p5.dance.js
```diff
@@ -1138,6 +1157,27 @@ module.exports = class DanceParty {
       title,
     };
 
+    let currentMeasure = Math.floor(this.getCurrentMeasure());
+    this.sprites_.forEach(sprite => {
+      if (sprite.alternatingMoveInfo) {
+        let alternatingMoveInfo = sprite.alternatingMoveInfo;
+        let val = Math.floor((currentMeasure - sprite.alternatingMoveInfo.start) / sprite.alternatingMoveInfo.cadence);
```
epeach: Nit - val isn't a very descriptive variable name.

src/p5.dance.js
```diff
@@ -1138,6 +1157,27 @@ module.exports = class DanceParty {
...
+        if (val % 2 == 0) {
+          if (sprite.alternatingMoveInfo.current != 1) {
+            this.changeMoveLR(sprite, sprite.alternatingMoveInfo.move1, -1);
+            sprite.alternatingMoveInfo = alternatingMoveInfo;
```
epeach: Why do we need to reset this?

src/p5.dance.js
jmkulwik: Could we ever end up with an off-by-one error? i.e. I'm imagining this scenario:
val = 1
val = 2
val = 3
val = 3
val = 5

src/p5.dance.js
epeach: I'm not sure I understand what this 2 does?

src/p5.dance.js
```diff
@@ -1138,6 +1157,27 @@ module.exports = class DanceParty {
...
+            sprite.alternatingMoveInfo.current = 1;
+          }
+        } else {
+          if (sprite.alternatingMoveInfo.current != 2) {
+            this.changeMoveLR(sprite, sprite.alternatingMoveInfo.move2, 1);
+            sprite.alternatingMoveInfo = alternatingMoveInfo;
+            sprite.alternatingMoveInfo.current = 2;
+          }
```
jmkulwik: What happens if none of these cases are met? Does the sprite skip a move?

epeach:
Did you test this against the cases you outlined on the whiteboard?

src/p5.dance.js
ajpal: `changeMoveLR` clears the `alternatingMoveInfo`, but we don't want to lose it in this case because we want to keep alternating

src/p5.dance.js
ajpal: `n` is used throughout the dance API to refer to measures, so I think it's okay to keep consistent with that pattern

src/p5.dance.js
ajpal: I don't think so (but I could be wrong so please double-check my logic here!)
You can calculate from start, cadence, and current what the dance _should_ be. But this is keeping track of what the current dance _is_. This is so that we can call `this.changeMoveLR` only when the dance needs to change (ie. when what the dance should be and what the dance is are not the same).
This check is happening inside the draw loop (roughly 50x per measure), so we don't want to call `this.changeMoveLR` every time, only when the move needs to change

src/p5.dance.js
ajpal: This is how I'm reducing the measure number to one bit so that we can alternate between two states.

src/p5.dance.js
ajpal: If none of the cases are met, the sprite continues doing whatever dance it was already doing.

ajpal:
image

ajpal:
> Did you test this against the cases you outlined on the whiteboard?

Yep! Table outlining those cases posted above. This implementation matches that behavior

src/p5.dance.js
ajpal: Synced offline- this is not possible. `start` and `cadence` never change, and `currentMeasure` increases at a constant rate, so the quotient will also increase at a constant rate.

PR #593, opened 2019-11-14
Add New Backgrounds to Random Chooser

epeach:
Adds new backgrounds and foregrounds to the constants file where the random-effect chooser pulls from.

Link https://codedotorg.atlassian.net/browse/STAR-845?utm_medium=referral-external&utm_source=jira-slack&utm_term=

### DIFF
--- a/test/unit/effectsTest.js
+++ b/test/unit/effectsTest.js
@@ -75,7 +75,7 @@ test('random background effect', async t => {
   nativeAPI.p5_.randomSeed(0);
 
   nativeAPI.setBackgroundEffect('rand');
-  t.equal(nativeAPI.world.bg_effect, 'fireworks');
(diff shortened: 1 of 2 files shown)


test/unit/effectsTest.js
```diff
@@ -75,7 +75,7 @@ test('random background effect', async t => {
   nativeAPI.p5_.randomSeed(0);
 
   nativeAPI.setBackgroundEffect('rand');
-  t.equal(nativeAPI.world.bg_effect, 'fireworks');
```
ajpal: Why does this change?

epeach: My understanding is that this changes because we seed randomness to return the same value everytime and adding additional values to the list changes the values in the randomness calculation.

ajpal: ah makes sense, thanks

PR #594, opened 2019-11-14
fix quads

ajpal:
There were two issues here- not resetting state and creating additional sprites

1. I added a reset function to the Effect and called it from `p5.dance.reset()` I think this is a pattern we should follow for all backgrounds that have state to clean up, rather than the current practice of doing that cleanup work at the beginning of the next `init()`

2. The `all dancers do move` block used `this.p5_.allSprites`, but this background creates 12 additional sprites (4 edges, 8 vertices). Then when we tried to change the animation for these sprites, it failed with:
image
Fixed by setting `isDancer` to `true` for all dance sprites when we create dancers, then filtering `this.p5_.allSprites` to only include dancer sprites for `all dancers do more`

Nov-14-2019 16-18-25


joshlory:
Rad, thanks for fixing!

screensaver

PR #599, opened 2019-12-10
ignore non-dancer sprites for replay

ajpal:
See also https://github.com/code-dot-org/dance-party/pull/594
The quads background makes additional sprites for the vertices and edges, but which are not dancers. This adds filter logic to ignore those sprites for the replay log.

### DIFF
--- a/test/unit/replayTest.js
+++ b/test/unit/replayTest.js
@@ -6,11 +6,16 @@ class mockP5 {
     this.allSprites = [];
   }
 
-  createSprite() {
+  createDanceSprite() {
     this.allSprites.push({
       getAnimationLabel: () => {},
       mirrorX: () => {},
+      isDancer: true
     });
+  }
+
+  createEffectsSprite() {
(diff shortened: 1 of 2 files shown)


test/unit/replayTest.js
```diff
@@ -21,10 +26,15 @@ test('replay', t => {
   t.equal(replayLog.getLog().length, 1);
   t.equal(replayLog.getLog()[0].sprites.length, 0);
 
-  p5Inst.createSprite();
+  p5Inst.createDanceSprite();
   replayLog.logFrame({p5: p5Inst});
   t.equal(replayLog.getLog().length, 2);
   t.equal(replayLog.getLog()[1].sprites.length, 1);
 
+  p5Inst.createEffectsSprite();
```
Hamms: nit: I'd love to either see a comment explaining what this is testing, or see this test broken up into a couple different tests with descriptive names

PR #617, opened 2022-01-11
Higher Power background

breville:
This adds a new background effect to Dance Party: Higher Power.

Note that `pngcrush` was used on the new sprite sheet, but it doesn't use the more efficient tile packing used in the other sprite sheets.

https://user-images.githubusercontent.com/2205926/149317279-2f5e84c5-5b7b-44c4-82a0-a0f1f451164b.mov


KylieModen (approved):
LGTM, thanks Brendan

demo.js
```diff
@@ -43,7 +43,7 @@ const runCode = async function () {
 }
 
 textareaCode.value = textareaCode.value || `var cat = makeNewDanceSprite("CAT", null, {x: 200, y: 200});
-setBackgroundEffectWithPalette("disco_ball", "rand");
+setBackgroundEffectWithPalette("higher_power", "rand");
```
ajpal, line 46: Do you want to check in this change or was this for development only?

src/constants.js
```diff
@@ -85,5 +85,16 @@ module.exports = {
     tropical: ['#eb6493', '#72d7fb', '#7efaaa', '#fffe5c', '#ee8633'],
     neon: ['#e035a1', '#a12dd3', '#58b0ed', '#75e847', '#fdf457'],
     rave: ['#000000', '#5b6770', '#c6cacd', '#e7e8ea', '#ffffff'],
+  },
+  HIGHER_POWER_COLORS: {
```
ajpal, line 89: nice way to handle the palette differences!

src/Effects.js
```diff
@@ -158,6 +163,77 @@ module.exports = class Effects {
...
+
+        let ctx = p5._renderer.drawingContext;
+
+        ctx.save();
+        let gradient = ctx.createLinearGradient(425, 425, 425, 0);
+        gradient.addColorStop(1, constants.HIGHER_POWER_COLORS[getCurrentPalette()][0]);
+        gradient.addColorStop(0, constants.HIGHER_POWER_COLORS[getCurrentPalette()][1]);
+        ctx.fillStyle = gradient;
+        ctx.fillRect(0, 0, 425, 425);
+        ctx.restore();
```
ajpal, line 201: out of curiosity: what does `save()` and `restore()` do?

ajpal (approved):

src/Effects.js
breville: I just borrowed this from similar code in the same file, but it turns out that this [saves](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/save) and then restores the state of the regular HTML5 [canvas](https://p5js.org/reference/#/p5/drawingContext).  I presume that in particular it restores the original `fillStyle`.

PR #618, opened 2022-01-12
Use Node 14

breville:
This updates `dance-party` to build with Node 14.

Notes:

- The `canvas` module had to be upgraded to build successfully with Node 14.
- The `jsdom` module had to be upgraded so that it had a canvas element.
- A number of visual baselines changed a little, but not significantly, presumably because of differences in the upgraded canvas engine used for the visual tests.  The actual tutorial's rendering should not be affected.


test/visual/backgrounds.js
```diff
@@ -71,7 +71,7 @@ async function testBackground(t, name, effect) {
   ['snowflakes'],
   ['sparkles'],
   ['pineapples'],
-  ['pizzas'],
+  //['pizzas'], error with Node 14 & Canvas 2.8.0: "node: cairo-arc.c:189: _cairo_arc_in_direction: Assertion `angle_max >= angle_min' failed."
```
aaronwaggener, line 74: Does this require further investigation? Or should we remove this line and the background?

aaronwaggener (commented):
Can you say more about why there are assets being added and removed as part of this upgrade?

test/visual/backgrounds.js
breville: This is a [known issue](https://nvd.nist.gov/vuln/detail/CVE-2019-6462) in Cairo, with a patch available [here](https://github.com/clearlinux-pkgs/cairo/blob/master/CVE-2019-6462.patch), but I haven't yet figured out a way to get this fix onto my system.  Both my own Ubuntu build and also that used by Github Actions (see https://github.com/code-dot-org/dance-party/pull/619) exhibit this issue.

We want to keep the background in the tutorial.  And I think it's actually okay to comment out this line, as several other lines in the same array have been similarly commented-out because of challenges with the visual testing.

aaronwaggener: Ah, I hadn't expanded the file to see this was following an existing pattern. Thank you for the explanation 👍

aaronwaggener (approved):

PR #646, opened 2023-09-13
hoc2023: initial work

breville:
This change updates the `dance-party` repo to build with the same key tooling as our main repo.  In particular, it now builds with `node` version `18.16.0`, uses `yarn` instead of `npm`, and uses `webpack` version `5`.

I believe that for Ubuntu builds, tests requires Ubuntu 20.04 or later, rather than 18.04 (which our main repo's drone docker image [still uses](https://github.com/code-dot-org/code-dot-org/blob/9b6ac19ef78a96171d4bd69b614c9d9ad0a144c4/docker/dockerfiles/Dockerfile#L8)), because the `canvas` component requires GLIBC 2.28 or later, and Ubuntu 18.04 only has GLIBC 2.27.  The GitHub Actions used to test PRs in this repo [currently get](https://github.com/code-dot-org/dance-party/pull/646/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R11) Ubuntu 22.04.

The trickiest part of this change was getting the `test:integration` component of the tests running again.  It uses [Karma](https://github.com/karma-runner/karma), which is now deprecated, but is at least now at the latest version.  A number of tricks were needed to make it work again with all of these latest versions.  I took a cursory look at a couple alternatives, but Karma offered the convenience of preprocessing with `webpack`, invocation of headless chrome, and does still work.


.github/workflows/build.yml
```diff
@@ -20,11 +20,11 @@ jobs:
         with:
           node-version: ${{ steps.nvm.outputs.NODE_VERSION }}
       - name: Install dependencies
-        run: npm ci
+        run: yarn
       - name: Build
-        run: npm run build
+        run: yarn run build
```
Hamms, line 25: Nit; I believe you can omit `run` from `yarn` commands; ie, `yarn dev` rather than `yarn run dev`. `yarn run` still works, though!

Hamms (approved):
This looks great!

Targeting a version of Ubuntu greater than 18 makes sense to me; we're planning to update our main repo to 20 soon enough, and I think testing against 22 should be just fine.

PR #650, opened 2023-09-17
hoc2023: initial AI

breville:
This pairs with https://github.com/code-dot-org/code-dot-org/pull/53769 in the main repo, which provides an initial AI block for Dance Party.  

Here, on the Dance Party side, the AI's response JSON is parsed and can be used to set the background, foreground, and possibly add a new dancer.

This is an early foundation for internal testing, and will soon change to match the proposed design.

### DIFF
--- a/src/p5.dance.js
+++ b/src/p5.dance.js
@@ -109,6 +109,7 @@ module.exports = class DanceParty {
     this.world.SPRITE_NAMES = constants.SPRITE_NAMES;
     this.world.MOVE_NAMES = constants.MOVE_NAMES;
 
+    // Store a reference to the main repo's AI function.
     this.doAi = doAi;
 
     if (spriteConfig) {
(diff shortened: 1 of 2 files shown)


src/p5.dance.js
```diff
@@ -1116,6 +1120,29 @@ module.exports = class DanceParty {
     }, this);
   }
 
+  // Called when executing the AI block.
```
fisher-alice, line 1123: Would it make sense to move this up where `setForegroundEffect`, `setBackgroundEffect` and `setBackground` are defined?

src/p5.dance.js
```diff
@@ -1116,6 +1120,29 @@ module.exports = class DanceParty {
...
+  async ai(value) {
```
fisher-alice, line 1124: Could we rename this function? Maybe something like `aiGenerateStage`?

src/p5.dance.js
```diff
@@ -49,6 +49,7 @@ module.exports = class DanceParty {
     i18n = {
       measure: () => 'Measure:',
     },
+    doAi,
```
fisher-alice, line 52: Could we rename `doAi` to something like `postOpenaiRequest

fisher-alice (commented):
Looks good! Just had a couple questions.

src/p5.dance.js
breville: This relates to the name of the black and the function it calls in the main repo [here](https://github.com/code-dot-org/code-dot-org/pull/53769/files#diff-539aa11323c94e9339a5d0316f6072470618ce8c67ef57acaf2d792ee87be8caR5).  I agree it should be something more specific, but it doesn't feel like our design is quite crisp enough yet.  As we discussed separately, let's address better naming in the next week as we do the next round of work.

src/p5.dance.js
breville: It might do, though it also remains to be seen whether we use those functions or call an all new system.  I think we should revisit once we make a bit more progress.

fisher-alice (approved):
Thanks for laying the groundwork!

PR #653, opened 2023-09-22
Assign default bpm value for sparkles background effect

fisher-alice:
This PR adds to the sparkles background effect a default value of 120 for the `bpm`. It is the only background effect whose `draw` function passes `bpm` as a parameter. The new 'preview' feature that will be added by https://github.com/code-dot-org/code-dot-org/pull/53817 calls on the `draw` nativeAPI function but does not load the selected song. Thus, we need to provide a default `bpm` value when the sparkles background effect is called on in preview mode.

There was a `testBackground` test failure for 'sparkles' after I added the default `bpm` value to the 'sparkle' draw function because the expected value for the test is 0. The test image is expecting a `bpm` value of 0. See [this commit](https://github.com/code-dot-org/dance-party/pull/653/commits/fce9fc040ac526165e0298d09735df7956f9757b) which confirms this. 
I deleted the current 'sparkles' baseline image, and a new one was created based on [this logic](https://github.com/code-dot-org/dance-party/blob/35cc8e7d9d771461ef443ff48f01934f72ab7f4c/test/visual/backgrounds.js#L23) that will be used for subsequent tests - test is now passing.

Here's a screencast video AFTER the update of the 'sparkles' background effect displayed in preview mode.

https://github.com/code-dot-org/dance-party/assets/107423305/05204f3a-8aee-418f-918e-e57633191f23


src/Effects.js
```diff
@@ -581,6 +581,7 @@ module.exports = class Effects {
       },
       update: function () {},
       draw: function ({bpm}) {
+        bpm = bpm || 120;
```
breville, line 584: Maybe worth a comment explaining why this needs a fallback?

fisher-alice: Added!

breville (approved):

PR #654, opened 2023-09-27
[HOC Dance Party] Add new background effects to Dance Party from Poetry

fisher-alice:
This PR adds 4 new background effects to Dance Party. I will not add these to the background dropdown for the 'set background effect` yet. However, we can add these 4 effects to the possible inputs for the new 'Use AI to' block.
The 4 background effects are:
- 'blooming_petals'
- 'clouds'
- 'frosted_grid'
- 'starburst'

These 4 were different enough from current Dance Lab effects to warrant adding them.

The first three are available as background effects in Poetry projects defined  in `backgroundEffects.js` [here](https://github.com/code-dot-org/code-dot-org/blob/381e9b93f7cbd081738dfa7adbc9e7ce4e169a0c/apps/src/p5lab/poetry/commands/backgroundEffects.js#L27), and the last one is a foreground effect in Poetry defined [here](https://github.com/code-dot-org/code-dot-org/blob/381e9b93f7cbd081738dfa7adbc9e7ce4e169a0c/apps/src/p5lab/poetry/commands/foregroundEffects.js#L235)
 in `foregroundEffects.js`. 

I had to some updating of the code when adding the background effects to `Effects.js` and a little more modification when updating the 'starburst' foreground effect.

After update: Note that in the screencast I changed the Dance Party level property `usesPreview` to `true` so that the user can view the background effect even when the program is not running.

https://github.com/code-dot-org/dance-party/assets/107423305/b8e392fd-b4a4-4edd-88aa-abe0baeab3ce

# Links
[jira](https://codedotorg.atlassian.net/browse/LABS-191)

# Testing
I updated a unit test and added a handful of visual tests for the background effects. I uncommented the tests for 'smiling_poop' and 'hearts_red' since `randomSeed(0)` is called [here](https://github.com/code-dot-org/dance-party/blob/main/test/visual/helpers/createBackgroundScreenshot.js#L16) in `createBackgroundScreenshot.js` before the background tests are run for consistent images.


src/Effects.js
```diff
@@ -485,6 +489,148 @@ module.exports = class Effects {
     this.ripples = createRipplesEffect(false);
     this.ripples_random = createRipplesEffect(true);
 
+    this.blooming_petals = {
```
breville, line 492: Maybe worth a comment on each of these new ones mentioning where they came from?

src/Effects.js
```diff
@@ -485,6 +489,148 @@ module.exports = class Effects {
...
+            ...color,
+          });
+        }
+      },
+      init: function () {
+        this.paletteLength = constants.PALETTES[getCurrentPalette()].length;
+        // initialize with enough petals to fill the screen - this is mostly
+        // useful so that preview shows what the background actually looks like.
+        // increment from 3 down to 0 so that petals are layered correctly with
+        // bigger petals behind smaller peals.
```
breville, line 510: nit: `petals`

super nit: capitalize sentences

test/visual/backgrounds.js
```diff
@@ -91,7 +91,10 @@ async function testBackground(t, name, effect) {
   ['ripples_random'],
   ['squiggles'],
   ['growing_stars'],
-  ['paint_drip']
+  ['paint_drip'],
+  ['blooming_petals', 'vintage'],
+  ['clouds', 'neon'],
+  // ['frosted_grid', 'vintage'], // anchor points are randomly generated, so this test will fail
```
breville, line 97: Is it worth having a test-specific code path to avoid randomness, so that we can get coverage on these?

src/shapes/frostedGrid.js
```diff
@@ -0,0 +1,41 @@
+module.exports = function drawFrostedGrid(p5, anchors, circles, spacing) {
```
breville, line 1: Similarly, a comment on each of these explaining that they are duplicates, and where they are copied from, could be helpful for future visitors.

test/visual/backgrounds.js
fisher-alice: Good suggestion! Yes - there is actually an already defined `randomNumber` function in `Effects.js` which uses `p5.random` instead of `Math.random`. This allows us to use `p5.randomSeed(0)` in the test before the background tests are run for consistent images. Thanks!

fisher-alice: I expanded test coverage of effects to include 'smiling-poop' and 'hearts_red' since they both used random integers as well.

breville (approved):
Thanks for seeing this all the way through, from analyzing the best effects to add, to porting them over, to finding a way to get test coverage.  I didn't review the effects themselves, but the work looks thorough, and it will be great to get these in use.

PR #655, opened 2023-09-28
hoc2023: AI dancers

breville:
This adds support for optional background dancers in the AI block, which is being added in https://github.com/code-dot-org/code-dot-org/pull/53983.

### DIFF
--- a/src/p5.dance.js
+++ b/src/p5.dance.js
@@ -1137,7 +1137,11 @@ module.exports = class DanceParty {
       if (
         params.dancers &&
         params.dancers.type &&
+        this.world.SPRITE_NAMES.indexOf(params.dancers.type.toUpperCase()) >=
+          0 &&
         params.dancers.count &&
+        params.dancers.count >= 0 &&
+        params.dancers.count <= 40 &&
         params.dancers.layout
       ) {
         this.makeNewDanceSpriteGroup(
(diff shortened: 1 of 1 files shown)


src/p5.dance.js
```diff
@@ -184,6 +184,11 @@ module.exports = class DanceParty {
   }
 
   async loadCostumeAnimations(costume, costumeData) {
+    if (!this.animations[costume]) {
+      // Invalid costume, nothing to do:
```
fisher-alice, line 188: Add a console statement similar to line 974.

src/p5.dance.js
```diff
@@ -966,7 +971,7 @@ module.exports = class DanceParty {
         sprite.rotation = 0;
       });
     } else {
-      throw new Error('Unexpected format: ' + format);
+      console.log('Unexpected layout format: ' + format);
```
fisher-alice, line 974: Just noting that if OpenAI responds with an invalid costume, we ignore and will not display dancers. However, the generated code and explanation will still reference the hallucinogenic response, in the example below, the costume 'rabbit'.
Screen Shot 2023-10-02 at 3 26 10 PM

fisher-alice (approved):
Nice addition!

src/p5.dance.js
breville: Yes, it's a good point.  I think we should work out a plan for detecting this situation and explaining it as a teachable moment!

src/p5.dance.js
breville: Done.

PR #656, opened 2023-10-10
Add AI block validation

fisher-alice:


src/p5.dance.js
```diff
@@ -81,6 +81,7 @@ module.exports = class DanceParty {
 
     this.world.bg_effect = null;
     this.world.fg_effect = null;
+    this.world.aiBlockSuccess = false; // This helps keeps track of whether the AI block successfully executed or not.
```
breville, line 84: nit: the rest of this file seems to use comments above the line in question, so I think best to do it that way here, too.

src/p5.dance.js
```diff
@@ -183,6 +184,28 @@ module.exports = class DanceParty {
     this.allSpritesLoaded = true;
   }
 
+  getUserCode(userCode) {
```
breville, line 187: I'd expect a "get" function to return a value.  This seems more like a "retrieve" or an "init" maybe?

src/p5.dance.js
```diff
@@ -183,6 +184,28 @@ module.exports = class DanceParty {
...
+    this.userCode = userCode;
+  }
+
+  getUserBlocks(userBlocks) {
```
breville, line 191: Same here.

src/p5.dance.js
```diff
@@ -290,6 +313,8 @@ module.exports = class DanceParty {
     this.allSpritesLoaded = false;
     this.songStartTime_ = 0;
     this.analysisPosition_ = 0;
+    this.world.aiBlockSuccess = false;
+    this.world.aiBlockNoParams = false;
```
breville, line 317: I find double-negatives difficult to work with, so am wondering if we could change this into `aiBlockHasParams` or similar?  

That said, the combination of `aiBlockSuccess` and `aiBlockNoParams` have four different possibilities, but I'm not sure we need all of them, so perhaps we could combine them into a single enum instead?

src/p5.dance.js
fisher-alice: Oh yeah - had already updated this. Thanks!

src/p5.dance.js
fisher-alice: I actually just removed this initialization.

src/p5.dance.js
fisher-alice: I'd like to keep these variables independent of one another so we can gather more detailed information for feedback purposes.

In the case of a level with more than one AI block, we initialize`aiBlockHasParams` to `true`. If one of the AI blocks doesn't have params defined, then the value of `aiBlockHasParams` should be set to `false`. Similar for `aiBlockSuccess`, the value is first set to `true` and if at least one fails for whatever reason, the value is set to `false`.

fisher-alice: But I did update the params variable to `aiBlockHasParams`. Thanks!

src/p5.dance.js
```diff
@@ -183,6 +183,46 @@ module.exports = class DanceParty {
...
+  // This function checks if the AI block is one of the children blocks of the given 'top' block.
+  // For example, if the given block is a 'When up pressed' event block, this function checks
+  // if the AI block is included within the event block.
+  isAiBlockChildOfTopBlock(topBlockType) {
+    let topBlock = this.userBlocksWithNextBlock.find(b => b.type === topBlockType);
+    while (topBlock.nextBlockType) {
+      if (topBlock.nextBlockType === 'Dancelab_ai') {
+        return true;
+      }
+      topBlock = this.userBlocksWithNextBlock.find(b => b.id == topBlock.nextBlockId);
```
bencodeorg, line 221: Does this represent a `topBlock` when set here? Or is it just we're iterating through each of the blocks below the top block initialized in the first line of this function? If so, might be more clear to initialize `topBlock` with the first child of the top block type being passed in, and rename to just `block` or something.

bencodeorg (commented):
One meta question here -- do we currently do any Blockly stuff in this repo? If not, looking at the internals of blocks feels a little unusual to to me. It's a little hazy in my head, but wondering if we could keep the block manipulation logic in code-dot-org, and register a way for the dance API to access blocks (similar to what we do here, when we pass in a validation function):

https://github.com/code-dot-org/code-dot-org/blob/f4896810ef156cbed830fab30f2aad0b88769022/apps/src/dance/Dance.js#L621-L623

I'd imagine then keeping the simple validation API functions (eg, `hasAiBlock`) here.

src/p5.dance.js
```diff
@@ -183,6 +183,46 @@ module.exports = class DanceParty {
     this.allSpritesLoaded = true;
   }
 
+  setUserBlocks(userBlocks) {
```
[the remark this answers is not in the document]
fisher-alice: Nice idea! We're typescriptifying our JS!

src/p5.dance.js
fisher-alice: Yes, `topBlock` does represent the to block when initialized, but then iterating through the blocks below it. I'll define a new variable called `block` for the iteration. Thanks!

fisher-alice: I was having a conversation with myself about this, so thanks for airing it!

fisher-alice:
> One meta question here -- do we currently do any Blockly stuff in this repo? If not, looking at the internals of blocks feels a little unusual to to me. It's a little hazy in my head, but wondering if we could keep the block manipulation logic in code-dot-org, and register a way for the dance API to access blocks (similar to what we do here, when we pass in a validation function):
> 
> https://github.com/code-dot-org/code-dot-org/blob/f4896810ef156cbed830fab30f2aad0b88769022/apps/src/dance/Dance.js#L621-L623
> 
> I'd imagine then keeping the simple validation API functions (eg, `hasAiBlock`) here.

Cool idea! Do you mind if I  log it as a follow-up item? I'd like to get this PR merged sooner than later so curriculum writers can start writing validation code in levels.

src/p5.dance.js
```diff
@@ -290,6 +330,15 @@ module.exports = class DanceParty {
     this.allSpritesLoaded = false;
     this.songStartTime_ = 0;
     this.analysisPosition_ = 0;
+
+    // If there are multiple AI blocks in a workspace, this value is set to `false`
+    // if any of the AI blocks do not execute successfully.
+    this.world.aiBlockSuccess = true;
```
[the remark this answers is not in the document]
bencodeorg: Maybe call it `aiBlockFailure` (default false), and set to true if you ever observe a failure?

I also wish I could come up with a more clear naming for what a "failure" is (it's something like "expected behavior of setting foregrounds and backgrounds not observed"), but can't come up with a snappy name :)

bencodeorg (approved):
LGTM! Agreed we can look into the "keep Blockly stuff in code-dot-org" idea later :)

src/p5.dance.js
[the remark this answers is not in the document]
fisher-alice: I like your `aiBlockFailure` suggestion! I was considering @breville 's comment about not liking to work with double negatives https://github.com/code-dot-org/dance-party/pull/656#discussion_r1353382966, but I think considering the scenario of more than one AI block in a program, I think I'll go with your suggestion. Thanks!

src/p5.dance.js
```diff
@@ -183,6 +183,48 @@ module.exports = class DanceParty {
...
+  // Note that the AI block could be disconnected and technically not part of the user code.
+  hasAiBlock() {
+    return this.getUserBlockTypes().includes('Dancelab_ai');
+  }
+
+  // This function checks if the AI block is one of the children blocks of the given 'top' block.
+  // For example, if the given block is a 'When up pressed' event block, this function checks
+  // if the AI block is included within the event block.
+  isAiBlockChildOfTopBlock(topBlockType) {
+    const topBlock = this.userBlocksWithNextBlock.find(b => b.type === topBlockType);
```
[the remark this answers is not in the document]
fisher-alice: Sure! Will update.

src/p5.dance.js
```diff
@@ -183,6 +183,53 @@ module.exports = class DanceParty {
...
   }
 
+  // This function receives the user's blocks from the main Blockly workspace and
+  // stores the following properties for each block:
+  // {type: string, id: string, nextBlockType: string, nextBlockId: string}
+  // The next block is the block that is connected to the bottom of the given block.
+  // If there is no next block, nextBlockType and nextBlockId are assigned ''.
+  // This function is used to check if a block is connected to another given block,
+  // For example, is the AI block connected to the 'setup' block?
+  setUserBlocksWithNextBlock(userBlocks) {
```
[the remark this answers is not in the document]
fisher-alice: Thanks! Refactored using `setFuncContext`. So nice that we can see if the AI block was called via an event block.

src/p5.dance.js
```diff
@@ -290,6 +301,12 @@ module.exports = class DanceParty {
     this.allSpritesLoaded = false;
     this.songStartTime_ = 0;
     this.analysisPosition_ = 0;
+    this.world.aiBlockCalled = false;
+    this.world.aiBlockContextUserEventKey = null;
+    // This value is set to `false` if any of the AI blocks in a user program
+    // are called without defined parameters.
+    this.world.aiBlockHasParams = true;
```
bencodeorg, line 308: I agree generally with @breville's comment that double negatives are confusing, but I think its more confusing to have state that is defaulted to something that not really true / is kind of irrelevant to most runs through dance party (ie, when ai blocks aren't used). I'd prefer to default to something like `aiBlockHasInvalidParams = false`

src/p5.dance.js
```diff
@@ -1179,6 +1189,12 @@ module.exports = class DanceParty {
     }
   }
 
+  setFuncContext(type, key) {
+    this.world.userEventTriggeredType = type;
```
bencodeorg, line 1193: I see this in the validation code doc, but unclear to me what it represents? Can you elaborate?

src/p5.dance.js
```diff
@@ -1126,7 +1147,11 @@ module.exports = class DanceParty {
 
   // Called when executing the AI block.
   ai(params) {
+    this.world.aiBlockCalled = true;
     console.log('handle AI:', params);
+    if (this.contextType === 'this.p5_.keyWentDown' && this.contextKey) {
```
bencodeorg, line 1152: Nit: move this string into a constant?

Also, reading thru the PR description, the value for `aiBlockContextUserEventKey` is up/down/left/right, right? Could you throw a comment in here noting that, or include something in the variable name that indicates that this key isn't like a "key as in key/value", but rather a "key as in key pressed"?

src/p5.dance.js
bencodeorg: Also wondering if this validation could fail sooner if you wanted it to (maybe we don't from a curriculum perspective)?

```
if (nativeAPI.getTime("measures") > 4) {
  // If the Down key is pressed, but the AI block was not called by the key event, then fail immediately.
  if (World.aiBlockContextUserEventKey !== 'down') {
    console.log('failed bc you did not include the AI block in the when Down key pressed block');
     console.log('userEventTriggeredType', World.userEventTriggeredType);
    nativeAPI.fail();
  }
}
```

src/p5.dance.js
fisher-alice: This makes senses in this particular case. I'll update! @breville

src/p5.dance.js
fisher-alice: > I see this in the validation code doc, but unclear to me what it represents? Can you elaborate?

Just removed from the validation code - was just a print statement for debugging purposes. And I forgot to remove it from `p5.dance.js` so just deleted. I was planning to use to keep track of the `key` pressed, but then realized we already had a set defined for this purpose! `this.world.keysPressed`. Thanks for the catch of the extraneous code!

src/p5.dance.js
fisher-alice: Good call! Just updated - thanks!

bencodeorg (approved):
LGTM!

breville (approved):
Fantastic.  Thanks for working closely with the curriculum team on these requirements, and for persevering through a variety of ideas to land on what feels quite elegant!

src/p5.dance.js
```diff
@@ -183,6 +183,17 @@ module.exports = class DanceParty {
     this.allSpritesLoaded = true;
   }
 
+  setUserBlocks(userBlocks) {
```
sanchitmalhotra126, line 186: Minor, but would it be better to make `userBlocks` a require parameter to `play()`? I worry that we might get into an inconsistent state if we call `play()` but forget to call `setUserBlocks`, but this way we're always passing in the latest set of blocks when running the program.

src/p5.dance.interpreted.js
```diff
@@ -123,7 +125,9 @@ function runUserEvents(events) {
   }
 
   function executeFuncs(item) {
+    setFuncContext(item.eventType, item.param);
```
sanchitmalhotra126, line 128: Would `setInputContext` be a clearer name for this?

sanchitmalhotra126 (approved):
Just catching up on PRs, so apologies if i'm missing any context! Noticed two small things but otherwise this looks great!

src/p5.dance.js
fisher-alice: > Also wondering if this validation could fail sooner if you wanted it to (maybe we don't from a curriculum perspective)?
> 
> ```
> if (nativeAPI.getTime("measures") > 4) {
>   // If the Down key is pressed, but the AI block was not called by the key event, then fail immediately.
>   if (World.aiBlockContextUserEventKey !== 'down') {
>     console.log('failed bc you did not include the AI block in the when Down key pressed block');
>      console.log('userEventTriggeredType', World.userEventTriggeredType);
>     nativeAPI.fail();
>   }
> }
> ```

I believe curriculum would want to give the user some time to press the 'Down' key while the program is running.

bencodeorg: Got it (I think), just finally wrapping my head around that you're not actually inspecting the blocks themselves now, so an immediate check isn't possible (right?)

src/p5.dance.interpreted.js
fisher-alice: After Slack discussion, we agreed to keep `setFuncContext`. Thanks!

src/p5.dance.js
fisher-alice: Omg - I just love this suggestions. Just updated - thanks!

src/p5.dance.js
fisher-alice: Yes - that's right!

PR #658, opened 2023-10-13
Add Poetry Lab Color Palettes to Dance Lab

fisher-alice:
This PR adds Poetry Lab color palettes to Dance Lab that can be used for the 'set background effect' block and the `setBackgroundEffectWithPaletteAI` to the API. This PR pairs with https://github.com/code-dot-org/code-dot-org/pull/54218.
Sam tweaked some of the poetry palettes so that the sprite dancers are more visible or 'pop' against the new background colors - [commit](https://github.com/code-dot-org/dance-party/pull/658/commits/7cdfc9cbc6e65e697c127bb8941ece609d56dbba).
I also added the 'starburst' to the list of background effects which was omitted from this past PR https://github.com/code-dot-org/dance-party/pull/654

Here's a screencast video of the

https://github.com/code-dot-org/dance-party/assets/107423305/9f2f1b5b-2170-454d-a116-9d0aee885e25

 new poetry colors added to Dance Lab:

Here's a screencast video fo the new poetry colors and effects with the AI block modal:

https://github.com/code-dot-org/dance-party/assets/107423305/7dce3474-0f89-47de-9377-65d3b38e0c23

# Links
[trello](https://trello.com/c/4qnud1Y9/51-make-poetry-background-palettes-available)


src/api.js
```diff
@@ -27,6 +27,9 @@ module.exports = class DanceAPI {
       setBackgroundEffectWithPalette: (effect, palette = 'default') => {
         nativeAPI.setBackgroundEffect(effect.toString(), palette.toString());
       },
+      setBackgroundEffectWithPaletteAI: (effect, palette = 'default') => {
```
jmkulwik, line 30: NIT: Add a comment here explaining why we have duplicate functions

fisher-alice: Good call - just added!

thomasoniii: I'm confused - why are there the two identical versions of this method?

The comment says that there are additional effects and palettes in the AI version, but that's just a change to the arguments that could use the same original method, right?

fisher-alice: Yes - good summary!

PR #660, opened 2023-10-19
Support "at x measures" block

bencodeorg:
Support a new "at X measures" block, which is only distinct from our existing "after X measures" block in that it runs right when we hit X measures (rather than after the measure, ie X + 1).

Open to suggestions on naming :)

Companion PR (with screenshots) in `code-dot-org` here: https://github.com/code-dot-org/code-dot-org/pull/54369

### DIFF
--- a/src/p5.dance.interpreted.js
+++ b/src/p5.dance.interpreted.js
@@ -178,14 +178,31 @@ function whenPeak(range, func) {
 }
 
 /**
+ * Execute code at a specified time, either in measures or seconds.
+ * One idiosyncrasy: despite the function's name, if we call atTimestamp(4, "measures", foo),
+ * we actually fire on the 5th measure (i.e after 4 measures have completed).
  * @param {number} timestamp
  * @param {string} unit - Should be "measures" or "seconds"
  * @param {function} func - Code to run when event fires
  */
 function atTimestamp(timestamp, unit, func) {
-  // Despite the functions name, if we call atTimestamp(4, "measures", foo), we
-  // actually want to fire on the 5th measure (i.e after 4 measures have completed)
-  if (unit === 'measures') {
+  atTimestampEventGenerator(timestamp, unit, func, true);
+}
+
+/**
+ * Execute code at a specified time, either in measures or seconds.
+ * Has "NotAfter" suffix to distinguish from "atTimestamp",
+ * which executes code **AFTER** x measures (if time is specified in measures).
+ * @param {number} timestamp
+ * @param {string} unit - Should be "measures" or "seconds"
+ * @param {function} func - Code to run when event fires
+ */
+function atTimestampNotAfter(timestamp, unit, func) {
+  atTimestampEventGenerator(timestamp, unit, func, false);
+}
+
+function atTimestampEventGenerator(timestamp, unit, func, afterMeasure) {
+  if (unit === 'measures' && afterMeasure) {
     timestamp += 1;
   }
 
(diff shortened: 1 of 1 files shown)


src/p5.dance.interpreted.js
```diff
@@ -178,14 +178,31 @@ function whenPeak(range, func) {
...
+
+/**
+ * Execute code at a specified time, either in measures or seconds.
+ * Has "NotAfter" suffix to distinguish from "atTimestamp",
+ * which executes code **AFTER** x measures (if time is specified in measures).
+ * @param {number} timestamp
+ * @param {string} unit - Should be "measures" or "seconds"
+ * @param {function} func - Code to run when event fires
+ */
+function atTimestampNotAfter(timestamp, unit, func) {
```
sanchitmalhotra126, line 200: Haha this is a tough one 🙂 I don't mind this name, maybe `atTimestampOrMeasure`/`onTimestampOrMeasure` as another idea?

sanchitmalhotra126 (approved):

src/p5.dance.interpreted.js
bencodeorg: When I did this yesterday I was thinking that I couldn't change the existing function name for legacy reasons, but realizing today I think that's not true (ie, we only serialize the name of the block, so we can update the block definition to call a different function and things will still work fine)? @fisher-alice can you confirm my understanding?

To align with the text that actually appears in the blocks, I could change the existing function to `afterTimestamp` and make the new one `atTimestamp`?

src/p5.dance.interpreted.js
```diff
@@ -178,14 +178,31 @@ function whenPeak(range, func) {
...
+ * we actually fire on the 5th measure (i.e after 4 measures have completed).
  * @param {number} timestamp
  * @param {string} unit - Should be "measures" or "seconds"
  * @param {function} func - Code to run when event fires
  */
 function atTimestamp(timestamp, unit, func) {
-  // Despite the functions name, if we call atTimestamp(4, "measures", foo), we
-  // actually want to fire on the 5th measure (i.e after 4 measures have completed)
-  if (unit === 'measures') {
+  atTimestampEventGenerator(timestamp, unit, func, true);
```
fisher-alice, line 189: If you don't end up changing the name of this function, could you add a short comment letting reader know what `true` is assigned?
Something along the lines of: 
```
// true is assigned to `afterMeasure` in following call
atTimestampEventGenerator(timestamp, unit, func, true);
```

src/p5.dance.interpreted.js
fisher-alice: Yes I believe that's correct. You could change the `func` property in the 'after x measures' block config file after updating the function name. That would be so much less confusing!

fisher-alice: I just looked and the name of the `json` file is `Dancelab_atTimeStamp.json` - so we wouldn't change the name of this file - is that correct? So still kinda confusing to a reader.

bencodeorg: Oh right, the block name can't be changed. So maybe just leave as-is 🤦

fisher-alice (approved):
I think the name you chose seems to work under the circumstances. I would lean towards making less changes to legacy dance blocks so I like the approach to took in this PR.

src/p5.dance.interpreted.js
bencodeorg: I added docs to the new eventGenerator helper (incl. a description of the afterMeasure parameter), hopefully that clarifies

fisher-alice: Perfect - even better!

PR #663, opened 2023-10-24
Dance AI: Live Preview

sanchitmalhotra126:
Enable live preview. Internally, this calls on `p5` to start looping and sets various flags and data related to song metadata. This is more or less just `play()` without song playback and validation callbacks, etc.

Pairs with https://github.com/code-dot-org/code-dot-org/pull/54466

https://github.com/code-dot-org/dance-party/assets/85528507/c69ab167-95be-4d36-b22f-a1ba8d710464


breville (approved):
Nice and simple!

PR #668, opened 2023-11-06
Loop animations in live preview

sanchitmalhotra126:
Allow animations to loop in live preview, as suggested in https://github.com/code-dot-org/code-dot-org/pull/54694. This allows us to keep the preview metadata JSON relatively small and lets live preview run for however long.

Tested locally with a variety of BG/FG effects.

### DIFF
--- a/src/p5.dance.js
+++ b/src/p5.dance.js
@@ -142,6 +142,9 @@ module.exports = class DanceParty {
 
     this.songStartTime_ = 0;
 
+    // Whether to loop analysis events. Used in live preview.
+    this.loopAnalysisEvents = false;
+
     new P5(p5Inst => {
       this.p5_ = p5Inst;
       this.resourceLoader_.initWithP5(p5Inst);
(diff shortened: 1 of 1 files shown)


fisher-alice (approved):
LGTM! Nice and simple.

breville (approved):

thomasoniii (approved):
lgtm 🚀

thomasoniii:
Not for this PR, but it looks like we could do some refactoring on the `constructor` and `reset` methods - there's a number of duplicated assignments (`loopAnalysisEvents`, `world.bg_effect`, `world.fg_effect`, etc) and I'd worry about them getting out of sync. Pulling them out into a common `setupDance` (or whatever) method would keep us DRY.

There are also some inconsistencies - like `aiBlocKHasInvalidParams` is set to `false` in `reset`, but uninitialized in the constructor, or `world.validationState` which is also reset but not initialized.

sanchitmalhotra126:
> Not for this PR, but it looks like we could do some refactoring on the `constructor` and `reset` methods - there's a number of duplicated assignments (`loopAnalysisEvents`, `world.bg_effect`, `world.fg_effect`, etc) and I'd worry about them getting out of sync. Pulling them out into a common `setupDance` (or whatever) method would keep us DRY.
> 
> There are also some inconsistencies - like `aiBlocKHasInvalidParams` is set to `false` in `reset`, but uninitialized in the constructor, or `world.validationState` which is also reset but not initialized.

Good call. I think this file and a few others (`Effects.js` for example) could use a good refactoring pass when we have time.

PR #669, opened 2023-11-06
Add world variables to support new validation support for AI Dance Party

fisher-alice:
This PR adds two new `world` variables to provide more validation support for Dance Party levels.

The curriculum team would like to give more targeted feedback for common user errors. The particular case that motivated the addition of this validation support is in the Dance Party progression, [level 2](https://studio.code.org/s/dance-ai-2023/lessons/1/levels/2). The user is asked to create a dancer (`makeAnonymousDanceSprite`), then have it do a new dance move (`changeMoveEachLR`). 

A common bug is when the two blocks (are out of order. Currently, the feedback is inaccurate because the curriculum team also wanted to address the other common error which is when the two blocks are in the correct order but they don't have matching costumes. Now we provide support for the case when the two blocks have the same costume but are out of order.
 
Before update:

https://github.com/code-dot-org/code-dot-org/assets/107423305/3de73440-ac3f-4fce-a8d9-7bcdf4cc2731

After update:

https://github.com/code-dot-org/code-dot-org/assets/107423305/c5496179-8af2-4199-91fd-7de6ea0ed042

### DIFF
--- a/src/p5.dance.js
+++ b/src/p5.dance.js
@@ -83,6 +83,9 @@ module.exports = class DanceParty {
 
     this.world.keysPressed = new Set();
 
+    this.world.spriteGroupsCalledToChangeMove = [];
+    this.world.spriteStyles = [];
+
     this.peakThisFrame_ = false;
     this.energy_ = 0;
     this.centroid_ = 0;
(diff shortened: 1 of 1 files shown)


src/p5.dance.js
```diff
@@ -83,6 +83,9 @@ module.exports = class DanceParty {
 
     this.world.keysPressed = new Set();
 
+    this.world.spriteGroupsCalledToChangeMove = [];
+    this.world.spriteStyles = [];
```
breville, line 87: I wonder if we should name/comment/group these kinds of variables so that future users of this code realize that they are related to validation, rather than used internally.

breville (approved):
Nice and simple!  Glad you were able to devise this solution, especially so close to launch.

src/p5.dance.js
fisher-alice: Good call. Added a comment for the ai validation variables as well.

PR #671, opened 2023-11-16
Handle invalid background/foreground effects

sanchitmalhotra126:
Handle invalid background/foreground effects & palettes being set. This is pretty unlikely to happen, but if it does then we'd hit a crash trying to call `init()` on `undefined`. This could potentially occur, for example, if we accidentally change one of the AI outputs to an invalid value. This change will also log a warning to a supplied logger if it exists, which will look like:
Screenshot 2023-11-16 at 3 30 51 PM

Trello: https://trello.com/c/qUhwOQHI
Pairs with: https://github.com/code-dot-org/code-dot-org/pull/54999

### DIFF
--- a/src/p5.dance.js
+++ b/src/p5.dance.js
@@ -46,6 +46,7 @@ module.exports = class DanceParty {
     showMeasureLabel = true,
     container,
     spriteConfig,
+    logger,
     i18n = {
       measure: () => 'Measure:',
     },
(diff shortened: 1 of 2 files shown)


cnbrenci:
Should we have some UTs to validate no crash?

sanchitmalhotra126:
> Should we have some UTs to validate no crash?

Good call! I kind of just assumed we had no unit testing on this code and would need to follow up later to add tests, but looks like we do

cnbrenci (approved):

PR #672, opened 2023-12-01
Break up Effects.js

bencodeorg:
A first step in refactoring our quite large Effects.js file (ie, the code that renders the details of each foreground/background). I put each effect into its own file, and organize into background and foreground effects folders.

I tried to make the changes in this PR as minimal as possible (ie, only moving code between files and including function arguments where needed), but a couple next steps I'm thinking about:
- some additional refactoring to make it so we don't have to pass as many helpers to each effect
- [use the foreground and background effects constants lists](https://github.com/code-dot-org/dance-party/blob/a7297ab74995d2095880effd6ada2458190f642e/src/constants.js#L46-L94) to enforce that each effect should have an associated file at `/effects/effectName`

## Testing story

I manually tested each background and foreground effect to see that it rendered as expected. We also have some unit testing on effects to make sure that a single frame looks as expected, which all passed.


sanchitmalhotra126 (approved):
Love it! Thanks for diving into this! I know the goal of this PR was to minimize code changes so I'm good with this as is, but I also had a few other follow-ups in mind:

- Since we're creating new files, can we go ahead and use ES6 import/exports? e.g. `export default function(p5...)` instead of `module.exports =` and `import` instead of `require`?
- Thoughts on creating classes for each effect? It would be a little more verbose, but could make parsing the code a little easier? e.g.
```
class BloomingPetals {
  constructor(p5, getCurrentPalette, colorFromPalette) {
    ...
  }

  init() { ... }
}
```
then in Effects.js, `this.blooming_petals = new BloomingPetals(p5, getCurrentPalette, colorFromPalette)`

fisher-alice:
> * Thoughts on creating classes for each effect? It would be a little more verbose, but could make parsing the code a little easier? e.g.
> 
> ```
> class BloomingPetals {
>   constructor(p5, getCurrentPalette, colorFromPalette) {
>     ...
>   }
> 
>   init() { ... }
> }
> ```
> 
> then in Effects.js, `this.blooming_petals = new BloomingPetals(p5, getCurrentPalette, colorFromPalette)`

@sanchitmalhotra126 , could you explain why this would be easier to parse?

fisher-alice:
> > @sanchitmalhotra126 , could you explain why this would be easier to parse?
> 
> It's just more modern syntax - the current format seems like it's the older pre-ES6 version of classes where you'd write it as a function that returns an object, but still use the `this` keyword to refer to properties and functions declared on the object. I personally find the class syntax a little clearer in displaying what the object properties and functions are, as well as what the `this` keyword refers to.
> 

Gotcha - so it would be easier for humans to read. 

> Also, thinking ahead to if we were to eventually start using TypeScript in this package, we could declare an interface/abstract class that these classes implement/inherit from. These could define the signatures for some common functions like `init` and `draw` for example.

Cool - thanks for the explanation!

bencodeorg:
@sanchitmalhotra126 agree re: classes, was just easier to copy/paste if I stuck to the existing older syntax.

Re: ES6 import/export, I played with using `import` / `export default` and see these errors, is there a babel plugin we're missing or something?

image


Both of these do feel like work items that could be picked up as standalone follow-ups.

bencodeorg:
Oh this is interesting re: Babel config when using `dance-party` consumed by `code-dot-org` (this is from `code-dot-org` config):

```
// Using babel.config.json instead of .babelrc makes this configuration apply
// even when we are transpiling within node_modules.
```

https://github.com/code-dot-org/code-dot-org/blob/staging-next/apps/babel.config.json#L1-L2

PR #673, opened 2023-12-01
Break up background and foreground effects

bencodeorg:
A proposed follow-up to https://github.com/code-dot-org/dance-party/pull/672 -- there is very little shared between background and foreground effects, so we could potentially totally separate them.

Some pros and cons I considered while putting this together -- I'm leaning towards this feeling like a good idea, but could be dissuaded :):

Pros:
- Clear delineation between foreground and background effects
- Palettes are only relevant to background effects, so we can keep palette-specific code with the background effects.
- Simpler constructor for foreground effects that reflects only the arguments it actually needs.

Cons:
- There's not much of (any?) difference between foreground and background effects from p5's perspective (ie, I think they're all just "effects" and our consumption of them is what breaks them into foreground and background effects), so it kind of makes sense to have them together
- There's at least one dependent repo (p5-replay, I think used to generate animation videos) that would need to be updated to accommodate this change. Refactor looks straightforward though (and we may have already made breaking changes that we haven't incorporated there yet).


test/unit/effectsTest.js
```diff
@@ -73,32 +74,33 @@ test('other background effects', async t => {
...
-  t.equal(nativeAPI.world.bg_effect, 'kaleidoscope');
+  t.equal(nativeAPI.world.bg_effect, 'circles');
   nativeAPI.getBackgroundEffect().draw({bpm: 120});
 
   t.end();
   nativeAPI.reset();
+  sinon.restore();
 });
 
-test('rainbow foreground effect updates with specified effect', async t => {
```
bencodeorg, line 86: This test is testing the `rainbow` effect, which is a background, not a foreground. `floating_rainbows` is the foreground rainbow effect.

bencodeorg:
Making a couple more changes here, will re-request review when they're done!

src/ForegroundEffects.js
```diff
@@ -0,0 +1,84 @@
...
+const confetti = require('./effects/foreground/confetti');
+const musicNotes = require('./effects/foreground/musicNotes');
+const paintDrip = require('./effects/foreground/paintDrip');
+const emojis = require('./effects/foreground/emojis');
+
+module.exports = class ForegroundEffects {
+  constructor(p5, alpha, getEffectsInPreviewMode) {
+    this.p5_ = p5;
+
+    // Duplicated in BackgroundEffects
```
bencodeorg, line 26: I ended up duplicating these helpers (rather than moving them into a utils file) because:
- p5.random is easier to stub in tests than Math.random, as it allows you to provide a random seed. Trying to stub Math.random using sinon ended up breaking some tests in ways I couldn't understand (JS out of memory issues?)
- some of the helpers (eg, `colorFromHue`) are defined with a default parameter that is passed in via the constructor (`alpha`), which made it a little hard to follow to define `colorFromHue` in a helper (ie, we needed a helper that would return a function). Seemed clearer just to duplicate.

src/constants.js
```diff
@@ -44,53 +44,53 @@ module.exports = {
   ],
   RANDOM_EFFECT_KEY: 'rand',
   BACKGROUND_EFFECTS: [
+    'blooming_petals',
```
bencodeorg, line 47: This is just alphabetizing the effects, no change to list.

src/p5.dance.js
```diff
@@ -59,6 +60,7 @@ module.exports = class DanceParty {
     this.showMeasureLabel = showMeasureLabel;
     this.i18n = i18n;
     this.resourceLoader_ = resourceLoader;
+    this.inPreviewMode = false;
```
fisher-alice, line 63: Did you need to add this initialalization? Or was it just for clarity?

src/BackgroundEffects.js
```diff
@@ -30,68 +30,30 @@ const growingStars = require('./effects/background/growingStars');
...
-const smileFace = require('./effects/foreground/smileFace');
-const confetti = require('./effects/foreground/confetti');
-const musicNotes = require('./effects/foreground/musicNotes');
-const paintDrip = require('./effects/foreground/paintDrip');
-const emojis = require('./effects/foreground/emojis');
-
-module.exports = class Effects {
-  constructor(p5, alpha, extraImages, blend, currentPalette = 'default') {
+module.exports = class BackgroundEffects {
+  constructor(p5, alpha, getEffectsInPreviewMode, extraImages) {
```
fisher-alice, line 34: If you keep foreground and background effects in different files, what do you think of not passing `alpha` as a parameter and defining it as a constant such as `COLOR_OPACITY` and referencing it in `colorFromHue`?

src/ForegroundEffects.js
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  // Duplicated in BackgroundEffects
+  /**
+   * Randomly pick one element out of an array.
+   * @param {Array.<T>} collection
+   * @returns {T}
+   * @private
+   */
+  sample_(collection) {
```
fisher-alice, line 81: I see that there is an underscore used for this function name since it's used only internally. I wonder if we should use this convention for other functions defined in this file that are only used internally as well. Or maybe remove this one so we're consistent?

src/BackgroundEffects.js
```diff
@@ -122,57 +83,23 @@ module.exports = class Effects {
...
+    this.diamonds = diamonds(p5, lerpColorFromPalette, getEffectsInPreviewMode);
     this.circles = circles(p5, lerpColorFromPalette);
     this.sparkles = sparkles(p5, randomColorFromPalette, randomNumber);
-    this.text = text(p5, lerpColorFromPalette, colorFromHue, randomNumber, getInPreviewMode);
+    this.text = text(p5, lerpColorFromPalette, colorFromHue, randomNumber, getEffectsInPreviewMode);
     this.splatter = splatter(p5, randomColorFromPalette);
     this.swirl = swirl(p5, randomColorFromPalette);
     this.spiral = spiral(p5, lerpColorFromPalette);
-    this.lasers = lasers(p5, lerpColorFromPalette, getInPreviewMode);
+    this.lasers = lasers(p5, lerpColorFromPalette, getEffectsInPreviewMode);
```
fisher-alice, line 93: I think you mentioned adding the user-facing names for effects as comments for easier reference?

fisher-alice (commented):
Nice work! I appreciate how you gave pros/cons of keeping effects together vs separated. After seeing these changes, I think I prefer having the foreground and background effects separated. I think if updates are made for a particular effect, it'll be easier to reference the effect.

Left a couple questions. Also curious if you were going to use the more modern syntax as @sanchitmalhotra126 had mentioned from https://github.com/code-dot-org/dance-party/pull/672#pullrequestreview-1760387150.

I love the clean up!

src/BackgroundEffects.js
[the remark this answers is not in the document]
bencodeorg: I like that! Will update.

src/p5.dance.js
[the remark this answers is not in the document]
bencodeorg: yeah mostly for clarity? It might be ok if I left it undefined, but seems more clear to default to `false`?

bencodeorg:
> Left a couple questions. Also curious if you were going to use the more modern syntax as @sanchitmalhotra126 had mentioned from [#672 (review)](https://github.com/code-dot-org/dance-party/pull/672#pullrequestreview-1760387150).

@fisher-alice re: modernizing syntax, I couldn't quite figure out why import statements aren't working, and it felt like work that could be picked up separately (maybe in conjunction with the "move effect definitions from object syntax to class syntax"?)

src/ForegroundEffects.js
fisher-alice: Interesting! Thanks for the explanation. I see this convention used in `p5.dance.js` as well.

fisher-alice:
> > Left a couple questions. Also curious if you were going to use the more modern syntax as @sanchitmalhotra126 had mentioned from [#672 (review)](https://github.com/code-dot-org/dance-party/pull/672#pullrequestreview-1760387150).
> 
> @fisher-alice re: modernizing syntax, I couldn't quite figure out why import statements aren't working, and it felt like work that could be picked up separately (maybe in conjunction with the "move effect definitions from object syntax to class syntax"?)

Sounds reasonable to me! Did you log a ticket for this?

test/unit/effects.js
```diff
@@ -46,3 +47,23 @@ test('Background effects', async t => {
   nativeAPI.reset();
   t.end();
 });
+
+test('All supported foreground effects exist', async t => {
```
fisher-alice, line 51: Thanks for adding these tests!

test/unit/effects.js
```diff
@@ -1,6 +1,7 @@
 const test = require('tape');
```
fisher-alice, line 1: How would you describe the difference between `test/unit/effects.js` vs `test/unit/effectsTest.js`? Would it make sense to combine these tests into one file?

fisher-alice (approved):
Looks good! Just left a question about a couple test files but I appreciate how much easier it is to read through the effects files and the test clean up.

test/unit/effects.js
bencodeorg: Doesn't seem like there's much difference to me, just combined them!

test/unit/effectsTest.js
```diff
@@ -201,3 +203,70 @@ test('logs invalid background palette', async t => {
   nativeAPI.reset();
   sinon.restore();
 });
+
+
+test('Set background effect to none, unknown, and known effects', async t => {
```
bencodeorg, line 208: This test was copied over from `effects.js` (which was deleted, as it seems redundant with this file)

PR #675, opened 2025-09-10
Add external renderer layer for generated dancers

mikeharv:
### Summary
Introduce a pluggable “external dancer” rendering layer that Dance Party can composite between sprites and foreground effects. This allows the main `code-dot-org` repo to supply its own renderer (e.g., a Lottie-based dancer preview) without adding that dependency to Dance Party.

The external renderer layer is shown below as bull dancer that is provided by a Lottie renderer introduced with https://github.com/code-dot-org/code-dot-org/pull/68250

https://github.com/user-attachments/assets/36846cb8-3235-4683-8da6-c183725a622e

We have multiple surfaces that need to render the same generated dancer JSON (Dance Party canvas, a Dancer Editor, and a Music Lab view). Centralizing the rendering logic in the main repo keeps Dance Party lean and lets other surfaces reuse the same renderer.

### What changed

- New adapter: `ExternalDancerLayer` (p5 adapter that owns a `p5.Graphics buffer` and hands its 2D context to an external renderer).
- New injection point: `DanceParty` now accepts an optional `externalRendererFactory` (constructor option). If provided, `setup()` creates an `ExternalDancerLayer`.
- New helper: `setExternalLayerSource(danceMove)` informs the render which dancer animation source JSON to use.
- Reset/teardown: Resets clear or dispose the external layer cleanly.

### Behavioral notes

- If no `externalRendererFactory` is provided, no behavior change: sprites and effects render as before.
- Timing for the external layer is driven by the current song measure, so it stays in sync with sprites.
- This code does not change any behavior for the standalone DanceParty package, unless a compatible external renderer is provided.

### Deployment strategy

Pairs with code-dot-org/code-dot-org#68250 (adds the concrete Lottie renderer and passes the factory into DanceParty).

### Testing story

- Verified dancer renders inside the Dance Party canvas and respects layout (fit/center/scale).
- Confirmed sprite split rendering shows “in front of”/“behind” correctly.
- Verified sync with song beat and sprite timing.
- Resizing and disposal don’t leak canvases.

### DIFF
--- a/src/GeneratedDancer.js
+++ b/src/GeneratedDancer.js
@@ -1,7 +1,7 @@
 // Thin p5 adapter: owns a p5.Graphics mid-layer and gives its 2D context
 // to the external renderer. CommonJS to match the rest of dance-party.
 
-// Map of move IDs to move names.
+// Map of move IDs to move names. These values are used to fetch source animation JSON.
 const movesById = {
   0: 'rest',
   1: 'clap_high',
--- a/src/GeneratedDancer.js
+++ b/src/GeneratedDancer.js
@@ -69,27 +69,14 @@ class GeneratedDancer {
 
   resize(worldW, worldH) {
     // detach old <canvas> to avoid DOM leaks
-    const el =
-      this.graphics &&
-      (this.graphics.elt ||
-        this.graphics.canvas ||
-        (this.graphics._renderer && this.graphics._renderer.canvas));
-    if (el && el.parentNode) {
-      el.parentNode.removeChild(el);
-    }
-    if (this.p5 && Array.isArray(this.p5._elements)) {
-      this.p5._elements = this.p5._elements.filter(e => e !== this.graphics);
-    }
-
+    this.dispose();
     this.graphics = this.p5.createGraphics(worldW, worldH);
     this.graphics.pixelDensity(1);
     this.renderer.init(this.graphics.drawingContext);
   }
 
   dispose() {
-    try {
-      this.renderer && this.renderer.dispose && this.renderer.dispose();
-    } catch (e) {}
+    this.renderer.dispose();
 
     const el =
       this.graphics &&
--- a/src/GeneratedDancer.js
+++ b/src/GeneratedDancer.js
@@ -1,10 +1,39 @@
 // Thin p5 adapter: owns a p5.Graphics mid-layer and gives its 2D context
 // to the external renderer. CommonJS to match the rest of dance-party.
 
-class ExternalDancerLayer {
+// Map of move IDs to move names.
+const movesById = {
+  0: 'rest',
+  1: 'clap_high',
+  2: 'clown',
+  3: 'dab',
+  4: 'double_jam',
+  5: 'drop',
+  6: 'floss',
+  7: 'fresh',
+  8: 'kick',
+  9: 'roll',
+  10: 'this_or_that',
+  11: 'thriller',
+  12: 'xarmsside',
+  13: 'xarmsup',
+  14: 'xjump',
+  15: 'xclapside',
+  16: 'xheadhips',
+  17: 'xhighkick',
+  18: 'xbend',
+  19: 'xfever',
+  20: 'xhop',
+  21: 'xknee',
+  22: 'xkneel',
+  23: 'xole',
+  24: 'xslide',
+};
+
+class GeneratedDancer {
   constructor(p5, worldW, worldH, renderer) {
     if (!p5 || !renderer) {
-      throw new Error('ExternalDancerLayer requires p5 and a renderer');
+      throw new Error('GeneratedDancer requires p5 and a renderer');
     }
     this.p5 = p5;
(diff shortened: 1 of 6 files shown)


src/p5.dance.js
```diff
@@ -1499,6 +1546,17 @@ module.exports = class DanceParty {
     console.warn(message);
   }
 
+  setExternalLayerSource(dancerName) {
+    if (!this.externalLayer) {
+      return;
+    }
+    const base = 'https://curriculum.code.org/media/musiclab/generate/dancers/';
```
breville, line 1553: Longer term, I think it would make sense for the main repo to know about asset URLs, with `dance-party` only having a concept of a dancer ID/metadata.

src/p5.dance.js
```diff
@@ -1554,7 +1612,43 @@ module.exports = class DanceParty {
       });
     }
 
-    this.p5_.drawSprites();
+    if (!this.externalLayer) {
+      this.p5_.drawSprites();
+    } else {
+      // Draw sprites in two passes, before and after the external layer.
```
breville, line 1618: Out of curiosity, would it be possible to create a regular `Sprite` but [override its `draw` function](https://p5play.org/docs/Sprite#draw) to call the custom renderer at the right time?

src/GeneratedDancer.js
```diff
@@ -0,0 +1,112 @@
...
+      this.graphics &&
+      (this.graphics.elt ||
+        this.graphics.canvas ||
+        (this.graphics._renderer && this.graphics._renderer.canvas));
+    if (el && el.parentNode) {
+      el.parentNode.removeChild(el);
+    }
+    if (this.p5 && Array.isArray(this.p5._elements)) {
+      this.p5._elements = this.p5._elements.filter(e => e !== this.graphics);
+    }
```
breville, line 104: Could this be shared code with what looks to be the same just above?

src/p5.dance.js
```diff
@@ -728,8 +856,14 @@ module.exports = class DanceParty {
     if (typeof group === 'object') {
       return group;
     }
+    if (group === 'GENERATED_DANCER' && this.generatedDancer) {
+      return this.p5_.allSprites.filter(sprite => sprite.isGenDancer);
```
breville, line 860: Can you use `getGeneratedDancerSprites`?

src/p5.dance.js
```diff
@@ -1263,12 +1402,22 @@ module.exports = class DanceParty {
...
+
+  getGeneratedDancerSprites() {
+    return this.p5_.allSprites.filter(sprite => sprite.isGenDancer);
+  }
 
+  getAdjustedSpriteDepth(sprite) {
+    let spriteScale = sprite.scale;
+    if (sprite.isGenDancer) {
+      spriteScale /= GENERATED_DANCER_SCALE;
+    }
```
breville, line 1416: To avoid a mutation, perhaps:
```
const spriteScale = sprite.isGenDancer
  ? sprite.scale / GENERATED_DANCER_SCALE
  : sprite.scale;
```

src/p5.dance.js
mikeharv: Great idea. Done.

src/GeneratedDancer.js
```diff
@@ -0,0 +1,99 @@
+// Thin p5 adapter: owns a p5.Graphics mid-layer and gives its 2D context
+// to the external renderer. CommonJS to match the rest of dance-party.
+
+// Map of move IDs to move names.
+const movesById = {
```
sanchitmalhotra126, line 5: Minor: share/consolidate with the list of names in https://github.com/code-dot-org/dance-party/blob/main/src/constants.js?

src/p5.dance.js
```diff
@@ -31,6 +32,7 @@ const WATCHED_RANGES = [0, 1, 2];
 
 const SIZE = constants.SIZE;
 const FRAMES = constants.FRAMES;
+const GENERATED_DANCER_SCALE = 0.75;
```
sanchitmalhotra126, line 35: For my understanding, why is this needed?

src/p5.dance.js
```diff
@@ -1499,6 +1648,14 @@ module.exports = class DanceParty {
     console.warn(message);
   }
 
+  setGeneratedDancerMove(source) {
```
sanchitmalhotra126, line 1651: nit: in a few places we call `this.generatedDancer.setSource` directly but we also have this function. Consolidate in favor of one or the other?

sanchitmalhotra126 (commented):
Nice! a few minor things/questions

src/p5.dance.js
mikeharv: Without it, the Lottie animation fills the entire canvas and the dancer looks quite a bit bigger than other sprites. At 75%, the size looks identical. I'm assuming that we scaled the original sprits down during the initial development in 2018 too, but I don't for sure. So admittedly, it's a bit of a magic number, but I'll add a comment.

PR #678, opened 2025-10-28
Mirror generated dancer sprites

mikeharv:
Relates to:
- https://github.com/code-dot-org/code-dot-org/pull/69142

This adds a way for generated dancer sprites to effectively mirror dance moves when needed. The `GeneratedDancer` layer will now manage a `mirror` property that gets updated during the sprite's `updateSpriteFrame` loop. When the animation has finished, we flip the `mirror` value. p5 handles the rest. This is only needed for dance moves that require mirroring.

With this change, generated dancers will be able to follow static sprite-sheet dancers across a sequence of any moves:

Before:

https://github.com/user-attachments/assets/354a7ece-10e0-4b5e-aa51-ed0768d776dc

After:

https://github.com/user-attachments/assets/316636e9-bae1-4c41-b313-a9412925a33d

### DIFF
--- a/src/GeneratedDancer.js
+++ b/src/GeneratedDancer.js
@@ -40,6 +40,8 @@ class GeneratedDancer {
 
     this.graphics = this.p5.createGraphics(worldW, worldH);
     this.graphics.pixelDensity(1);
+    this.shouldMirror = 1;
+    this.mirror = 1;
 
     // Hand the renderer our mid-layer 2D context.
     this.renderer.init(this.graphics.drawingContext);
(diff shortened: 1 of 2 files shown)


src/GeneratedDancer.js
```diff
@@ -40,6 +40,8 @@ class GeneratedDancer {
 
     this.graphics = this.p5.createGraphics(worldW, worldH);
     this.graphics.pixelDensity(1);
+    this.shouldMirror = 1;
+    this.mirror = 1;
```
breville: Could these be `true` instead of `1`?

mikeharv: I used `-1 | 1` to align to p5.play: https://code-dot-org.github.io/p5.play/docs/files/lib_p5.play.js.html

`.mirror` needs to stay that way but `.shouldMirror` could change to a boolean easily enough.

breville (approved):
Nice!

PR #679, opened 2025-11-15
Generated dancer: render quality

breville:
This improves the render quality of a generated dancer in dance party.

We've seen that the Lottie renderer does a nice job of rendering the generated dancer into a canvas, using the full resolution of the provided canvas to render nice anti-aliased shapes, as seen in the Generate Dancer levels of the Hour of AI 2025 progression.

Before now, the generated dancer in a dance party looked fairly pixelated, with noticeable aliasing artifacts.  On a retina display with a pixel density of 2, the dancer was first rendered into a 400x400 canvas.  This was then rendered into the final canvas, which is actually 800x800 pixels, at a 600x600 size.  

Now, we take advantage of the fact that the generated dancer is rendered in this progression at a consistent size to improve this pipeline a little.  We have Lottie render the dancer into a 600x600 canvas.  This is then rendered into the final 800x800 canvas, unchanged.

The result is a higher resolution dancer with somewhat fewer aliasing issues.

It's worth noting that the final 800x800 canvas is then scaled to the size of its element on the screen.  It would be interesting to always size this canvas to its onscreen size, but out of scope for now.

Thanks to @mikeharv for trying this out and sharing some more scenarios that this code could support if we introduce more blocks in the future.  As a result, it supports size (both "set to" and "change by") and position (initial, "set to", and "change by").

Tested with pixel density of 1, 2, and 3.

## Screenshots

Screenshots at pixel density of 2:

### Before

Screenshot 2025-11-14 at 7 14 50 PM

### After

Screenshot 2025-11-14 at 7 14 46 PM

### DIFF
--- a/src/p5.dance.js
+++ b/src/p5.dance.js
@@ -695,11 +695,7 @@ module.exports = class DanceParty {
         );
         this.p5_.push();
         this.p5_.scale(1 / this.p5_._pixelDensity);
-        this.p5_.image(
-          this.generatedDancer.graphics,
-          sprite.x - location.x,
-          sprite.y - location.y
-        );
+        this.p5_.image(this.generatedDancer.graphics, 0, 0);
         this.p5_.pop();
       }
     };
--- a/src/p5.dance.js
+++ b/src/p5.dance.js
@@ -693,7 +693,6 @@ module.exports = class DanceParty {
         sprite.mirrorX(
           this.generatedDancer.shouldMirror(this.getCurrentMeasure()) ? -1 : 1
         );
-
         this.p5_.image(
           this.generatedDancer.graphics,
           sprite.x - location.x,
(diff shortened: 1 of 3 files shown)


src/p5.dance.js
```diff
@@ -697,11 +693,10 @@ module.exports = class DanceParty {
         sprite.mirrorX(
           this.generatedDancer.shouldMirror(this.getCurrentMeasure()) ? -1 : 1
         );
-        this.p5_.image(
-          this.generatedDancer.graphics,
-          sprite.x - location.x,
-          sprite.y - location.y
```
breville, line 703: This was `0, 0` when the dancer remained in its starting location, or caused drift when it was moved.  Hardcoding to `0, 0` fixes the latter situation.

mikeharv (approved):
Nice fix! Great that we were actually able to simplify some of the sprite functions as well!

